PR #3, opened 2021-03-13
Added Conan Exiles template

eNcrypt65:
Windows only due to Funcom not making a Linux server


PhonicUK:
This doesn't look right, there are a few issues:

 - The RCON password shouldn't be a normal setting. You should use the ${{RemoteAdminPassword}} field and using UseRandomAdminPassword.
 - If the server accepts RCON for administration, is there not an RCON command that stops the server? If so the exit method should be 'string' with an ExitString of that command.

eNcrypt65:
Fixed the RCON issue, found that Conan actually has a non-writeable console thus ctrl-c is what the server software from Funcom uses.

eNcrypt65:
Fixed the RCON issue, found that Conan actually has a non-writeable console. ctrl-c is what the server software from Funcom uses.

PhonicUK:
If what you normally do is press CTRL+C, you should probably actually be using OS_CLOSE - can you try that and verify?

eNcrypt65:
Finally looked at this again. Didn't have a windows machine to test on. Tried all exit methods and only 'Kill' works now. Assume funcom update changed something

PhonicUK:
The setting in the manifest for MaxPlayers should use `$MaxUsers` as the FieldName to make it map to the MaxUsers setting properly. The settings should also have a DefaultValue rather than using the `CommandLineVars` setting (which can now be left empty)

Shamus03:
> Finally looked at this again. Didn't have a windows machine to test on. Tried all exit methods and only 'Kill' works now. Assume funcom update changed something

I have been using this template for a week or so with no issues.  Ability to configure some of the server settings via UI instead of the config files would of course be nice.

Only issue I've had so far is "restart" never seems to work (just stops the server and doesn't start it again) - I have a scheduled task to restart the server daily, and I had to add a second scheduled task about 15 minutes after that to make sure it got started back up.

eNcrypt65:
I'll look into the restart issue today.
I do plan on using AMPs ability to edit config files, just haven't yet

PR #4, opened 2021-05-20
Added basic Eco Server Module for AMP v2.1.1.4

stroskler:
Added basic Module for the game `Eco`

thankfull for any feedback. This configuration already works on my self-hosted AMP Instance


gOOvER:
There should be any Information, that the Server needs DotNet5 ;)

stroskler:
> 
> 
> There should be any Information, that the Server needs DotNet ;)

There are no additional needs since its installed with the steamcmd. I do not have to install additional packages. Could you clarify what exactly you mean?

stroskler:
> 
> 
> > > There should be any Information, that the Server needs DotNet ;)
> > 
> > 
> > There are no additional needs since its installed with the steamcmd. I do not have to install additional packages. Could you clarify what exactly you mean?
> 
> The Linux Version need DotNet on Server installed. Doesn Matter if it's installed by steam. DotNet5 is recommended. Mono is not longer supported since Eco 9.
> 
> I wrote and maintain the Egg for Pterodactyl Gamepanel, so i had some discussions with the Devs after Eco 9 release.
> 
> Info's about DotNet Linux: https://docs.microsoft.com/en-us/dotnet/core/install/linux

After re-checking the installed Ubuntu20 within the AMP configuration there is no `dotnet` installed. The only packages relating to microsoft are
```
ii  libgcab-1.0-0:amd64                 1.4-1                                 amd64        Microsoft Cabinet file manipulation library
ii  libmspack0:amd64                    0.10.1-2                              amd64        library for Microsoft compression formats (shared library)
ii  shim-signed                         1.40.4+15+1552672080.a4a1fbe-0ubuntu2 amd64        Secure Boot chain-loading bootloader (Microsoft-signed binary)
```

gOOvER:
The Net Runtimes are needed

stroskler:
Since this is not provided as readme and a fork of the `Template` i would deliver an additional  README or update the existing one. Any other suggestions?

PhonicUK:
Could you send this to http://github.com/CubeCoders/AMPTemplates instead please?

stroskler:
@PhonicUK correct me if i am wrong but it already is in the desired repository?

PhonicUK:
My bad! This is indeed the correct place. However the README should be omitted so it's purely a kvp and json - any extra steps should be listed in a Wiki article.

PhonicUK:
Also the command line flags aren't included properly. You're probably missing {{$FormattedArgs}}

stroskler:
if there are still missing options or configuration errors please let me know or advise me how to improve the module

PhonicUK:
* The world name is listed in your command line args, but has no presence in the JSON settings file.

Also can you as a sanity check validate that changing the settings in the user interface results in the expected change in-game?

PhonicUK:
Hey, any updates on this?

stroskler:
@PhonicUK there is still the problem, that the module is not well to configure. The most parameter will not work. It is only possible to spawn a single server instance. If you want to have to servers running beside manual changes needs to be done. I am already checking for updates on the server module from the publisher

PhonicUK:
What changes normally need to be made to make this behave?

stroskler:
To really change any game settings you need to edit manually the files which get generated after the initial server installation. They get created in `.ampinstances\[NAME]\Eco\SteamID\config\` the changes are permanent but cannot be templated or controlled initially modified AMP. The binary file has no parameter to set these settings. Maybe in future server releases more options are avaiable

PhonicUK:
I'll have to exclude it for now then since it won't be turnkey, what format is the config file? I'm looking at making Generic able to read/write config files.

stroskler:
the config directory looks like
image

and the config itself:
image

Maltazar:
I would very much like this template to, any progress in its making?

PhonicUK:
@stroskler Generic can produce config files these days, so you should be able to continue.

remy99c:
want to add to this that as of Aug 20 ECO supports RCON https://wiki.play.eco/en/RCON

PhonicUK:
Looks like they're using the Source RCON protocol which is nice to see.

stroskler:
Maybe in the near future I find the time to finish the addon . Sorry for the delay

PR #6, opened 2021-10-05
Avorion Dedicated Server

Shinynecrid:
Linux only for now. Windows to be added at a later date.


gOOvER:
sames as #7 ?

Shinynecrid:
> same as #7 ?

I submitted a bad first pull. It was fixed in 7

PR #9, opened 2021-11-06
Add BrokeProtocol Template

gOOvER:
Add BrokeProtocol Template


PhonicUK:
I wonder if this would run on Windows using WSL? I'll accept it as-is for now though.

gOOvER:
Sorry, i don't use AMP on Windows and can't test anything. I only create Servers for Linux

PR #10, opened 2021-11-06
Add Astroneer template

EkiciLP:


PhonicUK:
Have you tried running this using Proton? You can add Steam AppID 1580130 as an update stage that only applies to Linux.

PhonicUK:
Merging Windows only for now, but I'll try modifying it to accommodate Proton.

PR #14, opened 2021-12-22
Adding Sven Co-op

IceOfWraith:
This commit adds the ability to use Sven Co-op dedicated server utilizing the Generic Module.


IceOfWraith:
There was a bug with the regex. Undo!

PR #15, opened 2021-12-22
Add Sven Co-op

IceOfWraith:
This commit will add the Sven Co-op dedicated server utilizing the generic module.


xLeviNx:
Wow, you put a lot of work into this.

You don't think leaving a lot of these parameters to the server cfg file alone will suffice?

IceOfWraith:
That is always an option, but what would be the purpose of doing that? It seems AMP's goal is to make hosting easier by avoiding config files and providing a user interface.

-------- Original Message --------

PR #18, opened 2021-12-29
Add StarMade Module

IceOfWraith:
This adds StarMade Windows and Linux support. Currently all settings are managed in a single file with 188 settings. Once AutoMap receives new features to specify partial settings this will be added to the module as well.


PhonicUK:
Test comment to check discord integration.

PhonicUK:
Instead of calling it "Maximum RAM" - please refer to it as the Java heap size. Otherwise it can be misleading.

IceOfWraith:
Thanks for the recommendation. I've mirrored the other modules and submitted the change.

PR #25, opened 2022-01-08
Add Space Engineers

IceOfWraith:
This adds the config for the Space Engineers generic module including a base config file.


IceOfWraith:
I should note, this config will not work properly until this bug is addressed: [Checkbox InputType not working correctly in generic](https://github.com/CubeCoders/AMP/issues/536)

PR #26, opened 2022-01-09
Add tModLoader (draft)

jasperfirecai2:
Adds support for tModLoader using steamCMD. tModLoader64 can run on top of this but is not recommended due to its depedency on mono for linux, which would break AMP, and thus that kvp is not included


jasperfirecai2:
Draft to include updated checkboxes and beta release stream

PR #31, opened 2022-01-12
Update Space Engineers Generic to add Config Directory and Correct Checkboxes

IceOfWraith:
This commit adds the ConfigDirectory update step to Space Engineers generic template due to the directory not existing until the first server start.


IceOfWraith:
This should not be merged yet. I need to correct the checkbox values being reversed after the patch was put in.

IceOfWraith:
This one should be ready for publishing now. Sorry for any issues it's caused!

PR #32, opened 2022-01-14
Update Don't Starve Together Checkboxes

IceOfWraith:
Converted the checkboxes to match the latest format.


IceOfWraith:
I do not believe this one causes any issue for the game server. It is purely to clean up the code.

PR #39, opened 2022-01-15
OpenTTD

darkace1998:
This is a working Template

Done  :
KVP.
Config.json.
Meta.
Proper naming settings.
Config file on update command.

To Do :


IceOfWraith:
Have you addressed that there's no config file on first run? That will confuse most users.

darkace1998:
> Have you addressed that there's no config file on first run? That will confuse most users.

i have tryed to do that but it wil generate a hole directori tree so i dont see how i can make that

PhonicUK:
There's a 'CreateDirectory' update stage you can use.

PhonicUK:
The settings need proper display names and descriptions, using just the unmodified field name isn't helpful.

darkace1998:
Config download form github.
The settings have proper display names.

IceOfWraith:
In case someone wants to come along and finish this one. Here's the feedback I gave darkace1998.

1. The game is setup to only work on Linux, but looks like it can easily be converted to include Windows by adding the Windows executable line and setting the OS to 3. App.ExecutableWin=1536610\openttd.exe

2. Convert to the new ports system.

3. Add comma separated keywords. Usually this can just be the DisplayName with commas in between each word. This is what users can put in search to find the settings.

4. Missing a lot of descriptions. I'd recommend using this site, for most you can even just copy/paste exactly what they have. Please be sure to try and use the good capitalization and punctuation on the DisplayNames and Descriptions.
https://wiki.openttd.org/en/Archive/Manual/Settings/Openttd.cfg

6. If a setting is used for the command line variables you can add "SkipIfEmpty": true to that setting and it won't show at all if the user has no value there. I'm not sure how OpenTTD acts if you have a blank starting year, etc. But may be useful on this config.

7. I think "Sets the starting year." has an extra space in the FieldName/ParamFieldName.

Greelan:
Superseded by https://github.com/CubeCoders/AMPTemplates/pull/256

PR #41, opened 2022-01-17
Add Pavlov VR template 

Greelan:
Adding Pavlov VR dedicated server template.

Notes:

- Only Linux binaries are available for the dedicated server.
- The template relies on latest Nightly/CI builds after 2.3.2.0 for some functionality.
- The dependencies listed [here](http://wiki.pavlov-vr.com/index.php?title=Dedicated_server#System_setup_and_software_installation) need to be installed on the host.
- Port ranges 8177:8277 and 9100:9200 need to be added to AMP's default application port ranges for proper port allocation.
- The dedicated server does not have a writeable console.
- It does have RCON (telnet protocol), which the template configures.


PhonicUK:
Related issue: https://github.com/CubeCoders/AMP/issues/552

Greelan:
Related issue: https://github.com/CubeCoders/AMP/issues/562

IceOfWraith (approved):
This works as noted.

PhonicUK:
So does this no longer need relative port  mapping? What's the situation here?

IceOfWraith:
Yes, this game requires the base port and one 400 higher still. I only pushed it to complete because that's the same behavior as Arma 3 which does not currently assign relative ports either and is already implemented (I believe).

PhonicUK:
Okay so the issue is that even though you could set Port 1 to X and port 2 to X+400, you need AMP to be able to guarantee that relationship during deployments?

IceOfWraith:
That's right. For whatever reason they hard coded that the corresponding ports would be 400 apart. So 27000 works with 27400, 27001 works with 27401. Not sure the reasoning behind it, but it won't let you configure the second port separately.

PhonicUK:
Gotcha - so this can be merged in for now but needs AMP to have that ability to make it more reliable.

IceOfWraith:
Right, Greelan and I might have to play support a little extra until it's able to do that.

Greelan:
Given the second Pavlov port is not used by many if any other servers, practically it probably won't be an issue for this server. More likely an issue for Arma, DayZ, Unturned etc

At the very least though when new servers are added to AMP with non standard port ranges, can AMP automatically add a suitable port range to new instance defaults? Or get rid of the concept of application port ranges entirely...

This is gonna be more likely the issue for Pavlov with AMP - AMP will add a much higher second port because it doesn't have 8177 in the default range

PR #43, opened 2022-01-19
Add Wreckfest

Greelan:
Substantially completed template for Wreckfest Dedicated Server.

Notes:

- Minimum AMP version 2.3.2.
- Port range 33540:33640 needs to be added to AMP's default application port ranges for proper port allocation.
- DirectX dependencies may need to be installed on Windows. Despite having the latest DirectX installed, I also needed to add legacy libraries by installing `DirectX End-User Runtimes (June 2010)`.
- Currently there is no console input or output in AMP. As a result AMP cannot determine when the server is ready or when players join, leave or chat, despite the correct regex being included. Commands also cannot be sent to the server - instead the template currently immediately assumes the server is ready. If AMP's console can be fixed these aspects can be updated and improved.
- Essentially all settings have now been added to the template. Only event rotation settings still need to be configured manually in `server_config.cfg`.
- The server save data path is set to a `saved` directory in the datastore.


Greelan:
Related issue: https://github.com/CubeCoders/AMP/issues/578

Greelan:
This is probably as good as it gets. I've looked into whether the server console can be redirected to STDOUT so that AMP can read it, but that's not possible. As a result there is no console (and no tracking of server start or user joins and leaves). The user can refer to the server logs if needed.

Now also added support for Linux (Wine).

PR #44, opened 2022-01-22
Add Settings to Craftopia

IceOfWraith:
This adds all the settings to the Craftopia GUI and cleans up a few items.

The port must be put in manually until the functionality is added to use AMP variables in automap.


craftopia.kvp
```diff
@@ -1,34 +1,74 @@
...
-App.UpdateSource=Multi
-App.UpdateSources=[{"UpdateStageName": "SteamCMD Download","UpdateSourcePlatform": "All", "UpdateSource": "SteamCMD", "UpdateSourceData": "1670340"}]
-App.WorkingDir=1670340
+App.ExitFile=app_exit.lck
+App.HasWriteableConsole=True
+App.HasReadableConsole=True
+App.SupportsLiveSettingsChanges=False
+App.LiveSettingChangeCommandFormat=set {0} "{1}"
+App.ApplicationIPBinding=0.0.0.0
+App.ApplicationPort1=7778
```
Greelan, line 38: Should this be 6587 to match the config file (and the manifest)?

craftopiaconfig.json
```diff
@@ -1 +1,391 @@
...
+        "DisplayName": "Server Password",
+        "Category": "Server Settings",
+        "Description": "Set the server password, up to 8 digit number pin.",
+        "Keywords": "server password",
+        "FieldName": "serverPassword",
+        "InputType": "password",
+        "IsFlagArgument": false,
+        "ParamFieldName": "Host.serverPassword",
+        "IncludeInCommandLine": false,
+        "DefaultValue": "Password123",
```
Greelan, line 103: Should this be numerical?

craftopia.kvp
IceOfWraith: Yes, it should. I need to stop drinking while doing this. It's just water, but clearly I was incoherent.

Greelan:
> This template is affected by the config file [issue.](https://github.com/CubeCoders/AMP/issues/556)

A workaround is to download the template as is (say to the InstanceDir, so it sits alongside the other template files), and then run a CopyFilePath update stage to copy it to the working directory with the right name. This is what I have done for Pavlov VR, for example

IceOfWraith:
Thank you! I had thought of that as a workaround, but also wanted to bring awareness to how many configs are affected.

Greelan:
Agree the behaviour should be fixed!

PR #46, opened 2022-01-22
Add tModLoader module

jasperfirecai2:
Adds a tModLoader module each for the stable and alpha release branches. Stable is available from steam or github, 1.4 alpha is steam-only.
The 1.4 alpha is a different module because the binaries are entirely different and thus didn't fit without reasonable configuration using conditional updates.


jasperfirecai2:
This replaces #26 so that the fork branches are correct.

IceOfWraith:
Related issue: https://github.com/CubeCoders/AMP/issues/592

PR #49, opened 2022-01-22
Add Empyrion Galactic Survival

IceOfWraith:
This is a mostly working template. The config file needs setup to download from GitHub as well as the settings need created in the GUI.


IceOfWraith:
This config is affected by the limited number of ports available, 5 are required to enable Telnet/console. 4 are needed for the base game, so until that time this will not have a console output.

This is also affected by the lack of the [automap AMP variables feature](https://github.com/CubeCoders/AMP/issues/550).

PhonicUK:
Gonna hold this one for now, I really want to add a better method for arbitrary number of ports in Generic instances.

IceOfWraith:
Related issue: CubeCoders/AMP#582

IceOfWraith:
Ideally, this would use port ranges. if users add port range 30000:30100 they should be just fine. No chat regex since I didn't see a way to chat in game. Otherwise, no issues found.

IceOfWraith:
I somehow butchered this PR somehow. Making a new one somehow.

PR #50, opened 2022-01-24
Add Unturned template

Greelan:
This adds Unturned 3 Dedicated Server.

Note:
- The template ensures the server is started and stopped upon any update, which is important for installation and for any subsequent change to the server name.
- Minimum AMP version is 2.3.2.
- A number of settings are included in the UI (under “Server Settings”) for the purpose of populating `Commands.dat` (and for setting the server type). The user is able to add additional settings in `Commands.dat` manually. Some settings (those that use only a key, with no value), cannot currently be set through the UI given limitations in AMP and the requirements of Unturned for `Commands.dat`.
- Console input does not work through STDIO, but is able to occur via RCON. The template ensures that RCON is installed and configured.
- The server name is changeable. However, after changing it, the server must be updated (as noted above).
- The template assumes two config files will be included in the CubeCoders repo for downloading (for the main config file and for the workshop download file).


PhonicUK:
There's new methods to start the server as part of the update process, although these need further esting.

Greelan:
I saw that, but (a) the build with those settings seems to have failed so I can't test them anyway, and (b) the StartApplication stage seems to run every time an update is run, which is not really the desired behaviour. Usually it's only needed for first server start.

The template works perfectly fine within the current parameters of AMP so I am not sure why you have reverted this PR to draft?

Greelan:
This template is now in the best state that I think is possible for Unturned.

Greelan:
One other comment on the shutdown command. If this is issued from AMP's console, the server seems to do a proper shutdown (and of course is then immediately restarted by AMP). However the same command that is issued as an exit String from the kvp does not behave the same way.

I have noticed the same thing with Conan Exile - an "exit" command from the console does a proper shutdown, but as a String command it just seems to close RCON and then the server is killed.

Can this be looked into? I would have thought the String command to shut down the server through the kvp should be issued in the same way as a console command.

Shinynecrid:
that same shutdown issue is happening with Insurgency Sandstorm. if you type quit into the console line, it shut's down but when stop button is clicked on amp. nothing... it just waits the 30s timeout time before killing the process.

PhonicUK:
I'm a bit dumb - I configured it to close the RCON before it sends the exit command :D

Greelan:
Derp lol. Looks like the build with the fix has failed though. Once this is in and working, Unturned can be updated and should be done :)

Greelan:
And then hopefully I never have to look at that stupid server ever again 😂

Greelan:
Now updated to reflect the fix for the String ExitMethod. Template complete!

IceOfWraith (approved):
The config seems to be working great! I was able to quickly start the server, load in, verify the status tab,, and shut it down with no issues on Windows.

IceOfWraith:
This fixes #64

IceOfWraith:
Related issue: https://github.com/CubeCoders/AMP/issues/562

PR #53, opened 2022-01-29
PZ: various updates

Greelan:
- Bump minimum AMP version
- Add numerous new UI settings
- Configure RCON (AMP does not use it)
- Allow for different number formats in regex


PhonicUK:
Why is this using an initial password instead of the Random password option?

Greelan:
User convenience, so the user can set a static password without needing to worry about setting up RCON Passthru in AMP.

If you have an objection to the template setting a static password (that the user can change in the UI), I can remove it. (Of course, nothing to stop the user simply editing GenericModule.kvp to change the behaviour anyway...) Let me know?

Curious though as to why the option of setting a non-random password is allowed in the kvp if it's not the desired approach?

PhonicUK:
You should set the field type to RandomPassword for passwords. It's not so much that it's 'allowed' its just that 'password' is a normal HTML field type so it works.

Greelan:
Sorry, I'm dense. Can you explain further what you are after/suggesting?

PhonicUK:
`"InputType": " RandomPassword"`

Greelan:
Done. That is a nice feature (just tested it to see what it does).

PhonicUK:
Surely they'd just copy+paste it?

Greelan:
Dunno, but why does it matter? The default is no server password anyway

Greelan:
When `SkipIfEmpty` hits mainline, what's the likely AMP version number? I'd like to pre-empt that in this template for the min AMP version to avoid any issues

IceOfWraith (approved):
I approve of these changes. Nice work!

PR #55, opened 2022-01-29
Add Conan Exiles (draft)

IceOfWraith:
This brings back the work that was started on https://github.com/CubeCoders/AMPTemplates/pull/3 previously. It has been migrated to use the new generic settings. It currently only works on Windows, but has a baseline started for Proton. Settings have not been added to the GUI yet.


IceOfWraith:
This config is plagued by the [RCON timeout issue](https://github.com/CubeCoders/AMP/issues/575). This functions as expected otherwise. Proton support was added using xvfb. The following is needed in order to take advantage of Linux Conan servers:
`apt-get install xvfb lib32gcc1`

Greelan (approved):
This is getting to a good state. 

Would be ideal if the Required field in the config manifest could be implemented so the default admin password can be removed. (The password can't be empty when the server is started, so that's why currently a default is set.)

Would also be great if there was a way to run it in Proton that didn't require xvfb, as it adds about 500MB.

PhonicUK:
Related issue: https://github.com/CubeCoders/AMP/issues/575

cechacek:
https://discord.com/channels/266012086423912458/671396391804731405/951929213890469959 50USD bounty for "usable" conan exiles generic proton module

cechacek:
> https://discord.com/channels/266012086423912458/671396391804731405/951929213890469959 50USD bounty for "usable" conan exiles generic proton module

bounty fullfiled by Ice thank you. @PhonicUK i hope that RCON issue will be solved sooon so it will be ready for prime time :)

IceOfWraith:
Turns out this game has a hardcoded port that is the Game Port +1 so this is related: https://github.com/CubeCoders/AMP/issues/562

IceOfWraith:
Updated the command line args due to formatting issues with the game, added Server Password to the command line since it goes in the Engine.ini and is not a line included by default. Disabled RCON for now due to the bug with AMP disconnecting after some time. The server works fine with regular console but will not accept input.

IceOfWraith:
I added the QueryPort to the command line args, added blank EnumValues by default, and tested like crazy. I did notice the server browser can take a VERY long time (20 minutes sometimes) to show the server, but direct connect works from the start.

Greelan:
Comments from testing on Windows:

- We've discussed changing `$RemoteAdminPassword` to random and hiding the setting in the UI.
- We've also discussed having hidden manifest settings that deal with the ports in the relevant config files (to ensure that any user additions are controlled by AMP). This will also require additional metaconfig settings for the relevant files (eg `Engine.ini`).
- The template currently uses STDIO but my recollection is that RCON is needed for console input (but it has the timeout issue - see above).
- Sleep is configured (but SleepMode is not turned on by default?) but in my testing it did not work.
- Most importantly, the descriptions for some settings end in a period, and others do not. :)
- I like what you did with CustomMap.

Testing still to occur on Linux - because you know I *love* Proton...

IceOfWraith:
As always, I appreciate the thorough feedback! Here's my responses below:

-     $RemoteAdminPassword set to random and hidden from UI.
-     Hidden manifest settings added to deal with the ports.
-     SourceRCON is configured thanks to your suggestion with how to handle the password. Timeout issue does persist, but most settings, if not all, are managed in the in-game menu so I don't see this as a show stopper, just a big inconvenience.
-     Sleep was disabled by default because it doesn't work due to what I believe is a bug with AMP. I've re-enabled it since it doesn't hurt anything other than give false hope to users that the server will sleep. Refer to this issue for details: https://github.com/CubeCoders/AMP/issues/587
-     Someone else made the config manifest. I've cleaned up the absolutely TERRIBLE formatting. :p
-     Thank you! Gotta accommodate our modding community if possible.

Please let me know if you see any issues!

Greelan:
Nice. I don't you need a starting RemoteAdminPassword in the kvp - just leave it blank.

Also, shouldn't RconPassword in the config manifest have an ini header in the ParamFieldName?

IceOfWraith:
I left it just as a fallback. But yes, it's not needed. Adjusted the setting for RCON Password as well.

Greelan:
Good stuff. FWIW if sleep does not work, I would have it disabled. Avoids unnecessary support questions.

IceOfWraith:
I mainly kept it on for when a fix eventually comes along. I didn't want users to have to replace their instance or edit the kvp file to get it working. Once AMP is fixed it will just "work" again.

PhonicUK (commented):
RemoteAdminPassword should have its value cleared out before a pull request :)

Greelan:
Ice has already removed it based on my comment a couple of days ago

PR #67, opened 2022-02-13
Convert Insurgency Sandstorm to base config

IceOfWraith:
This switches the config to use a base config rather than a template to allow users to manually add special settings without a full config file overwrite.


Shinynecrid (approved):
I tested the new Auto Map. I was able to add additional lines to the config file as expected. Also tested map rotations, and switching maps/modes on the fly. I would say Changes approved on this one.

PR #68, opened 2022-02-13
Astroneer: Various updates

Greelan:
- Switch to automap
- Add manifest settings
- Fix Proton (even though not currently usable with server)


IceOfWraith:
I will test this today and get back to you.

IceOfWraith (approved):
We discussed through DM about the extra decimal places in the frame rate settings. This is the default value directly from the config.

We also discussed removing the password and whitelist checkbox, because the first screen in the game when you join as admin has you select those. They are pre-filled if you put them in AMP though, so we agreed to leave them for security purposes in case the admin doesn't connect immediately.

PR #71, opened 2022-02-14
Unturned: Fix two input types

Greelan:
Inadvertently reversed the input types for Steam64 ID and GSLT token. Thanks to @IceOfWraith for the pickup!


IceOfWraith (approved):
This works nicely now. :)

PR #72, opened 2022-02-14
Update Unturned steamclient.so method

IceOfWraith:
Changed Unturned to use the new environmental variable method rather than moving files.


Greelan (approved):
Thanks for the sleuthing on this. In my testing I can't detect a difference in the server operation - startup logs appear the same. I haven't tested with Maps.

Great if this means at least a couple of update stages can be removed!

PR #74, opened 2022-02-16
Fixed Craftopia Config

IceOfWraith:
Set the ConfigFormat to use quotes around all settings. I believe this was causing games with spaces in the name to be non-joinable. The strange part is you could set it back to having a space and it would work going forward.

Also set the ExitMethod to String and quit. The user must type yes following since there is no option to do two input commands to quit a server. CtrlC doesn't work in AMP on Windows for this game. And CtrlC doesn't work for this game on the Linux Terminal. OS_CLOSE ended in a kill.

Removed the config file and used the start/wait/stop update stages instead.


IceOfWraith:
Putting this as a comment so it's very clear what's going on with this one:
I believe that this game's dedicated server is the cause of the timeout issue. It is still in an early beta stage at this point. I have tried to recreate the problem and no matter what, I cannot now. After the first initial connection to any server it works from then on. The ini file is intended to not have quotes in this case given the examples at the top. There are other settings I've put in the GUI but do not seem to function in the server yet, things like "No Food", "No Death", etc. have no impact. You will die and you will be angry. Please contact the developer to have them fix their game.

PR #75, opened 2022-02-17
Create Stationeers Config

IceOfWraith:
This is the initial Stationeers generic config. The server functions well overall. The GUI settings appear to work as intended. I do not own this game and as such cannot test fully. I need someone to login and obtain the join/leave/chat console output so I can complete this one. Otherwise, I believe it should be finished.

Known issues:
The shutdown command does not work in Windows, only Linux using OS_Close. The server only allows commands to be put in through the web RCON otherwise.
Sleep mode fails to wake up: Related issue - https://github.com/CubeCoders/AMP/issues/587


IceOfWraith:
It turns out shutdown/sleep issues are due to running AMP as a service in Windows. This is not an issue with this config specifically.

IceOfWraith:
This was tested with the help of TK3062 on Discord. We did find direct IP connections do not seem to work but appears to be a game limitation. The server browser connected fine each time. Sleep mode is still an issue on Windows as a service. I've disabled sleep by default, but it is available for the user to enable.

PR #79, opened 2022-02-18
Add Killing Floor 2

IceOfWraith:
This is the beginning of the Killing Floor 2 config. It needs further settings added before release, but currently runs and can set the starting map.

Known issues:
The console does not connect as this starts a separate process: https://github.com/CubeCoders/AMP/issues/576
Cannot add all ports needed due to 3 port limitation: https://github.com/CubeCoders/AMP/issues/582


IceOfWraith:
I've determined the guide on the wiki is incorrect. There is no NTP port requirement for incoming connections (which makes sense), only external. So all ports are able to be assigned properly.

IceOfWraith:
This one is ready to test.

Created the settings needed for the metaconfig file. Added web management.  Only server start regex included due to only having player join regex otherwise. This would cause the player count to be off and there is no chat message either.

Console still has no output in Windows based on the above referenced issue.

IceOfWraith:
This was tested by FenixVale on Discord. Merging for now with the understanding that the console has no output in Windows as long as the bug persists.

PR #80, opened 2022-02-19
Unturned: update telnet login format

Greelan:
Needs to be specifically set to cater for new default


IceOfWraith (approved):
No issues found.

PR #87, opened 2022-02-23
Pavlov VR: Add server versions

Greelan:
In addition to default (public), can now select PC Beta, Shack Live or Shack RC


IceOfWraith (approved):
Works like a charm!

PR #88, opened 2022-02-24
Correct Stationeers Config

IceOfWraith:
This game has many bugs still and the dedicated server does not function properly. It is unable to create a new world. You must create a singleplayer/multiplayer game and copy it into the save folder. The line "OriginWorldSaveLocation" must be removed from the world_meta.xml file when importing.

-Remove uneccessary -basedirectory launch arg
-Removed -modpath launch arg that does not work. It must be manually set in the default.ini file
-Removed "" around launch args
-Corrected -worldname to -loadworld
-Clarified Map select option
-Set autosaveinterval to default at 300 per the default in the game


Greelan (approved):
Nothing jumps out at me with the changes. Good job

PR #89, opened 2022-02-26
Add The Isle (Legacy) template

Greelan:
Adds working The Isle (Legacy) Dedicated Server.

Notes:
- Console input does not work.
- Shutdown methods (such as OS_CLOSE, CtrlC and WM_CLOSE) do not work on Windows. AMP just kills the server after the timeout.
- Includes Proton support for Linux, however that is lacking, as AMP monitors and operates on the Proton process, rather than the underlying server process - so for example metrics are inaccurate, and while OS_CLOSE works, it does not shut down the underlying server process. Only stopping the instance entirely does that.
- Player join, leave and chat regex to be added.


Greelan:
Connected issue: https://github.com/CubeCoders/AMP/issues/576

PR #45, opened 2022-01-22
Add Merged Terraria module

jasperfirecai2:
Adds a generic terraria module that has more configuration options than the builtin one. see [this repo](https://github.com/jasperfirecai2/amp-templates/tree/main/Terraria) for dev version

PR #90, opened 2022-03-02
Add tModLoader module

jasperfirecai2:
Adds a standalone tModLoader module with all its dependencies. Will be replaced by https://github.com/CubeCoders/AMPTemplates/pull/45 if its feature request ends up being implemented


Shinynecrid:
Console doesn't function on windows. To look into further.

jasperfirecai2:
Discovered issue with files not saving to the right folder. Still needs to be looked into further

jasperfirecai2:
Make sure to manually grab the configtemplate when teesting, or to change the update stage to use my fork, since it will otherwise fail to template merge. WIndows console still does not work, so only autosaves and process-killing are available.

PR #91, opened 2022-03-02
Add Terraria module

jasperfirecai2:
Adds a standalone Terraria module with all its dependencies. Will be replaced by https://github.com/CubeCoders/AMPTemplates/pull/45 if its feature request ends up being implemented


Shinynecrid:
Console doesn't function on windows. To look into further.

jasperfirecai2:
Make sure to manually grab the configtemplate when teesting, or to change the update stage to use my fork, since it will otherwise fail to template merge. WIndows console still does not work, so only autosaves and process-killing are available.

IceOfWraith (changes requested):
Here's my feedback so far:
1. Add the following lines to your config so it doesn't waste ports.
`App.ApplicationPort2=0`
`App.ApplicationPort3=0`
`App.RemoteAdminPort=0`
2. The template names need to be lowercase to match the files. Fails in Linux:
`Meta.MetaConfigManifest=terrariametaconfig.json`
`Meta.ConfigRoot=terraria.kvp`
`Meta.ConfigManifest=terrariaconfig.json`
3. Linux doesn't appear to be making a world save file. The World folder is there, but no file saves under it.

Overall, it is very clean! The first two are minor things to fix, the third I'm not sure what's causing yet.

IceOfWraith (approved):
#3 was user error on my part. This all checks out after 1 and 2 were updated. Merging now!

PR #92, opened 2022-03-05
Add The Isle (Legacy)

Greelan:
Adds working The Isle (Legacy) Dedicated Server.

Notes:

- Console input does not work.
- Includes Proton configuration, but given that the functionality is lacking (eg in terms of AMP not monitoring and operating on the game server process), Linux support is not enabled by default.
- No user join or leave regex is included because there is no server output for that.
- Config file includes a number of settings that are also set in the UI, but there is no reliable information as to whether they are actually functional for the server.

Connected issues:
- https://github.com/CubeCoders/AMP/issues/587
- https://github.com/CubeCoders/AMP/issues/576


Greelan:
OS_CLOSE resolved in 2.3.2.8

IceOfWraith (approved):
Everything looks good here. Was able to join the game easily and adjust settings. Save appeared to work properly. Great work! Go ahead and adjust the download location and this should be merged.

PR #93, opened 2022-03-05
Add Mordhau Config

IceOfWraith:
This is the first draft of the Mordhau config. The current issues outstanding:
1. Console/RCON is not accepting input
2. Shutdown (OS_CLOSE) command does not work cleanly in Windows.


IceOfWraith:
This one is ready for primetime. It does not shutdown cleanly in Windows still, but that does not impact the server's ability to function normally. It does not store any data like an MMO would, so nothing is lost with a kill command.

PhonicUK:
Can you check against the new CI that should ship in 20 mins or so?

IceOfWraith:
Definitely!

PR #99, opened 2022-03-15
Plutonium BO2 Template

msmcpeake:


IceOfWraith:
I can't recall where you left off with this. Is there a legitimate way to obtain the files needed to use this server in AMP? Right now the config just says "redacted", but I thought there was an option we discussed of pulling the needed files from Steam.

IceOfWraith:
Closing due to lack of activity.

PR #100, opened 2022-03-16
Update Space Engineers

IceOfWraith:
There is a second config file I did not notice. Spaceengineers-Dedicated.cfg is used for the initial world creation, and some of the other global settings. Then, the Sandbox_config.sbc file is used inside the world folder for subsequent changes.


IceOfWraith:
I want to investigate this one further. I'm not certain the sbc file should be edited manually versus using the "ignorelastsession" setting. I've reached out on their Discord for clarification.

IceOfWraith:
Ignorelastsession is for battle sessions where progress is not saved. I also noticed it is needed if you change the world name or it will overwrite your old save file if using multiple. It's best to keep worlds saved in different instances, though. Both the .cfg and .sbc file should stay in sync per their Discord. This has no known issues at this time and is ready for testing.

PR #110, opened 2022-04-26
SCP: Secret Laboratory Dedicated Server

kysxe:
This is a WIP, but is functional if you enjoy browsing through the AMP user's "~/.config" folder. I'll probably have to wait until IORedir is released before it becomes fully working without hassle. 

I do also plan on working on the Windows version of the server, but I'd rather work on the Linux version first as oddly enough it's easier to deal with imo.


dhkatz:
Would love to see this available for wide use 😊

Greelan:
I started to play around with this template but the key issue is that the server writes its data outside the datastore, and there is not way (so far as I can find) to force a different location

gkuba:
If we ever get the IORedirect to make this functional here is a good source for config options
https://scpsl-faq.com/en/server-faq/server-configuration

Also something to note is for the server to show up in the server browser you have to go though a "verificiation" process with the devs.

NateStanley:
> If we ever get the IORedirect to make this functional here is a good source for config options https://scpsl-faq.com/en/server-faq/server-configuration
> 
> Also something to note is for the server to show up in the server browser you have to go though a "verificiation" process with the devs.

I know this isn't like, the "vanilla" server, but if you use MultiAdmin you can set a different directory for where the server looks for "appdata" as well as the config.

https://github.com/ServerMod/MultiAdmin

warden161:
Wiki configs is outdated, watch out btw.

For configs, you can change path using -configs startup argument.

not sure about plugins though

dhkatz:
This actually appears to be possible with the base game using LocalAdmin.

LocalAdmin supports the `--config` and `--logs` arguments (as well as `--gameLogs` to point to the server binary's logs)

You can also passthrough arguments to the server binary as well through LocalAdmin. Took a peek at MultiAdmin's source and it passs the `-logfile`, `-configpath` and `-appdatapath` to the server binary to change its locations.

warden161:
-appdatapath would be what's needed then.

PhonicUK:
What's this currently waiting on?

IceOfWraith:
I'm not sure it's waiting on you anymore. The IOREDIR isn't working last I knew, but they've found a work around that should work. Just needs someone to implement it.

Greelan:
I will dust this off and see if it can be finished.

smyers119:
> I will dust this off and see if it can be finished.

There is a couple problems with this template that keep it from being implemented as a generic template.  I have a template @smyers119/amp-scpsl , but there is 2 major work arounds that keep it from adding.
  
  - You need to use a custom docker image, I built off of ampbase and added mono-complete.
  - You need to create a empty folder for config files, I did this using FetchURL unzipping a empty folder.  
  

In my opinion this pull request should be closed, as there is no current way to make this as a generic within the set guidelines.

IceOfWraith:
Those are easily fixed. There is cubecoders/ampbase:mono for the first. And the second There's a CreateDirectory update stage that isn't documented.

smyers119:
> Those are easily fixed. There is cubecoders/ampbase:mono for the first. And the second There's a CreateDirectory update stage that isn't documented.

Awesome, is there a way to mark it docker only as well?  What happens if they try to run it without docker?

Greelan:
Lol, I've actually overhauled the template and added a bunch of stuff, it is now working without issue on Linux, including that nothing is tired outside the datastore. Just got to deal with one remaining file storage aspect outside the datastore on windows

smyers119:
> 



> Lol, I've actually overhauled the template and added a bunch of stuff, it is now working without issue on Linux, including that nothing is tired outside the datastore. Just got to deal with one remaining file storage aspect outside the datastore on windows

Nice, I was getting ready to add configuration pages, but that looks tedious so I am more then happy to have someone take over.

Greelan:
Settings are indeed the most tedious lol. In this case there are far too many (and appears mostly undocumented) so it will be more a matter of adding the basics and then anything else can be by request. The template is set up so the configs are in place from installation and AMP is able to write to them

Greelan:
I've also been toying with avoiding mono (given it can cause issues with AMP) and having the template supply it's own version on .NET, like I've done with other templates. My understanding is that mono/.NET is needed for plugins (the server itself runs fine without it). Anyone know whether using .NET as an alternative would be OK? Any known plugins that require mono which I could test with?

Greelan:
Closing as the source repo appears to have been deleted.

In any event, updated version in `Greelan/AMPTemplates:dev`. Some settings added, more likely to come. Currently not working in Docker due to an AMP bug.

Feedback welcome, would also appreciated log lines for user join/leave/chat.

smyers119:
I have it working fine in docker.

Greelan:
Coz you are not using the $ExternalIP var in AMP to automatically set the public IP in the config 😀

Greelan:
For testing purposes, I've taken out the $ExternalIP setting for the time being, until it is back working again. This means the template will work in docker now

Greelan:
The other thing to note is that given LocalAdmin is a precursor executable and in turn starts the actual server binary, metrics monitoring is not correct in AMP (child process monitoring on Linux does not help). AMP monitors LocalAdmin, not the actual server binary. So it shows incorrect memory usage, for example.

If LocalAdmin could be bypassed, that would be address this. I can see in htop the arguments it passes to the server binary, but without really any information available on those arguments (eg what console or id is for), it is difficult to decide how to bypass it. I've raised a ticket for info on the arguments in the official discord, but we will see if I get any info. It's possible that LocalAdmin does other things that make it not feasible to bypass it anyway.

smyers119:
You shouldn't be passing Appdatapath and configpath to the server binary.  As you see in LocalAdmin code they are platform based
```
            {
                CorrectPathFound = true;

                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                    return path + Path.DirectorySeparatorChar + ".config" + Path.DirectorySeparatorChar;

                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                    return path + Path.DirectorySeparatorChar + "AppData" + Path.DirectorySeparatorChar + "Roaming" + Path.DirectorySeparatorChar;

                ConsoleUtil.WriteLine("Failed to get special folder path - unsupported platform!", ConsoleColor.Red);
                throw new PlatformNotSupportedException();
            }
```

smyers119:
Can you put in the pull request so we can have this conversation over the actual pull request.  Also if you look at my smyers119/ampscpsl repo it has the regex for join and leave

smyers119:
I think I interpreted the above wrong.

Greelan:
I will create my PR. I will also add the regex - I thought you were waiting on that too.

Can you explain the last comment more? My aim in all this is to keep as much as possible (desirably everything) in the instance datastore, rather than just following what the server does outside AMP. I have achieved that through various ways on Linux, but Windows is not as co-operative (typical).

Greelan:
@smyers119 can you also share the log lines in which you based your regex?

PR #112, opened 2022-05-01
Add 7 Days to Die Generic

IceOfWraith:
This is the draft of the 7 Days to Die generic config. This is not thoroughly tested yet, and not tried on Linux at all. I believe most everything works as expected including UMOD in Windows though.

Known Issues:
1. No Telnet connection due to related feature request: https://github.com/CubeCoders/AMP/issues/623
2. No FPS Status - This is not currently possible with Generic


Greelan:
I realise this is a work in progress, but suggest adding the new `"DeleteAfterExtract":true` setting to the uMod download stages, so the zip is deleted after being extracted.

IceOfWraith:
The telnet issue is fixed from above.
There is no FPS tracking still, but I feel this should be released without that feature for now due to the change in config files from old versions causing issues for existing servers that can't be upgraded. This also accounts for both user join messages so old and new versions of the game will track players properly.

PR #114, opened 2022-05-03
Add Hurtworld

Greelan:
Initial template for Hurtworld dedicated server.

Notes:

- Ports 12871-12971 need to be added to AMP's default application port ranges for proper port allocation.
- The server does not appear to have any console on either Windows or Linux. Hence the template assumes the server starts immediately, and no regex is included for server start or user join/leave.
- uMod installation option included (credit to @IceOfWraith). Not yet fully tested.


IceOfWraith (approved):
Nice work on this one finding the right settings!

PR #120, opened 2022-05-19
Fix risk-of-rain-2 user join/leave detection

DataBitz:
Updated regex for current log entries.
e.g. 
User Join is
SteamworksServerManager.OnAuthChange(steamId=11111197969451111, ownerId=11111197969451111, status=OK)
User Leave is
Log: connection {1} has been disconnected by timeout; address {::ffff:192.168.1.111:54586} time {126493}, last rec time {124158} rtt {13} timeout {2000}
Attempt to send to not connected connection {1}


PhonicUK (commented):
There's no userid on the leave, so AMP won't know who left.

DataBitz:
> 

Correct, there is only a connection id on leave and not sure if that is much use unless AMP captured the connection on join.

PhonicUK:
AMP actually has some internal logic for using things like connection IDs but it's not exposed to Generic. Using the connection ID as the userid will work and AMP will do what you expect, but ideally the userid should be the steamID and I'll expose the session ID tag.

DataBitz:
> AMP actually has some internal logic for using things like connection IDs but it's not exposed to Generic. Using the connection ID as the userid will work and AMP will do what you expect, but ideally the userid should be the steamID and I'll expose the session ID tag.

Ok I'll have a go at making the userid the connectionid and user the steamid so at least the steamid is available for notifications etc.

PhonicUK:
I've added another capture group to the nightly build of AMP called `sessionid` which does what you're after :)

IceOfWraith:
I see the following messages:
`NetworkManagerSystem.AddPlayerInternal(conn=hostId: 0 connectionId: 1 isReady: False channel count: 8, playerControllerId=0, extraMessageReader=NetBuf sz:259 pos:259 `
`Log: connection {1} has been disconnected by timeout; address {::ffff:192.168.0.59:58013} time {741305}, last rec time {739105} rtt {12} timeout {2000}`
`SteamworksServerManager.OnAuthChange(steamId=76561197977084314, ownerId=76561197977084314, status=OK): Could not find connection for steamId. `

So we would use the sessionid from the first two and not have it linked to anything since the steamId is on another line?

IceOfWraith:
Looks like that worked like a charm. Doesn't give any player names/options but it updates the player count just fine! I'll include it in my Docker updated config.

DataBitz:
> 



> > AMP actually has some internal logic for using things like connection IDs but it's not exposed to Generic. Using the connection ID as the userid will work and AMP will do what you expect, but ideally the userid should be the steamID and I'll expose the session ID tag.
> 
> Ok I'll have a go at making the userid the connectionid and user the steamid so at least the steamid is available for notifications etc.

Yeah it looks like AMP will only scan a single log entry with RegEx at a time so we either have the SteamID or the SessionID.
The active player count does seem to be working for me despite AMP in theory not knowing exactly which player left.
This message works in posts to Discord and includes the SteamID "{@UserID} has joined AMP Server - ROR2"
So yeah player ID is there but not player name.

PR #124, opened 2022-05-21
Eco: Update beta branch selection config

Greelan:
More robust UI settings for beta branch selection (mirroring approach introduced in Project Zomboid)


Greelan:
Testing shows working as expected

PR #125, opened 2022-05-22
Initial Commit - V-Rising

tempusthales:
- No chat Regex
- No Linux at this time, but its coming. Ball is on Dev's court.


Greelan:
This appears to function as intended (although I haven't tested connecting with the game client). Couple of comments:

- I assume the `Immediate` start is deliberate because there is no relevant console output to trigger the ready state (despite the template including some regex)?
- I note the example server start batch file included with the server has the command line argument `-logFile ".\logs\VRisingServer.log"`. Has that been excluded deliberately?
- The Descriptions in the settings manifest have a number of inconsistencies with capitalisation and full stops/periods (and I would probably use different language for the names and descriptions). But I guess this is more personal preference/OCD in nature...

Greelan:
Oh, as suggested previously, I would also include suffixes (eg `seconds`) and placeholders where possible

tempusthales:
> - I assume the `Immediate` start is deliberate because there is no relevant console output to trigger the ready state (despite the template including some regex)?

Correct.

> - I note the example server start batch file included with the server has the command line argument `-logFile ".\logs\VRisingServer.log"`. Has that been excluded deliberately?

Yes.  If you ad the argument `-logFile ".\logs\VRisingServer.log"` then you can't turn on console logging, which to me was more important to have.  Maybe this is a bug on their side or a Unity problem.

> - The Descriptions in the settings manifest have a number of inconsistencies with capitalisation and full stops/periods (and I would probably use different language for the names and descriptions). But I guess this is more personal preference/OCD in nature...

I wanted to use the same names as in the original .json file to make it easy to troubleshoot.

tempusthales:
> Oh, as suggested previously, I would also include suffixes (eg `seconds`) and placeholders where possible

Will review and add if it's important.

gOOvER:
you can try to Run the Server in Proton. in wine, it runs fine

IceOfWraith:
I've made some initial attempts to use WINE. Proton failed completely. WINE only worked if using the virtual C drive rather than the native Linux file system. That also lined up with the docker images I've seen. If you have a method that doesn't require that it very likely could work in AMP.

PhonicUK:
Proton seems to work as per Ice's recent contribution. I also added the suffixes/etc.

PR #133, opened 2022-05-27
Update Risk of Rain 2

IceOfWraith:
This updates the following:

1. Runs on WINE
2. Docker capable!
3. Game mode option
4. Updates player join/leave regex using sessionid
5. New ports method


DataBitz:
See this message on docker and application not starting.
[00:58:15] [API:admin Activity] ??: Starting the application.
[00:58:15] [Generic:admin Info] ??: Required system-level tool '/usr/bin/xvfb-run' does not exist. Aborting startup.

Also, player count on Windows with the sessionid+nightly doesn't seem to be working. Had 3 players and either 0 or 1 player count.

IceOfWraith:
@DataBitz verify that you're using Docker to run it. I don't see any way it couldn't be working for you as the Docker container is what it is. If it works for one it should work for all.

The player count is very flakey. But it seems to be due to the game server. When you're in game a lot of times it even lists player names as "Unknown". Unfortunately, that's not something AMP can control. I'll verify I didn't miss anything on the regex for those but it's definitely not working right in the game server.

IceOfWraith:
Just to update, we're working through this on Discord. It seems the new files weren't loaded yet. Just didn't want anyone reviewing this for merging and thinking there was an issue. There might be. But nothing obvious yet. :)

DataBitz:
> Just to update, we're working through this on Discord. It seems the new files weren't loaded yet. Just didn't want anyone reviewing this for merging and thinking there was an issue. There might be. But nothing obvious yet. :)

Thanks for your help @IceOfWraith, didn't find the root cause but tested ROR2 with docker ok on Ubuntu 20.04 & 22.04 with clean installs.
Something weird with my Ubuntu 20.04 VM didn't allow it to work there.
Error from docker container was
A crash has been intercepted by the crash handler. For call stack and other details, see the latest crash report generated in:
 * C:/users/amp/Temp/Hopoo Games, LLC/Risk of Rain 2/Crashes
 0037:err:seh:setup_exception stack overflow 1680 bytes in thread 0037 eip 000000007bcd8c3e esp 0000000000130f80 stack 0x130000-0x131000-0x230000

IceOfWraith:
I'm going to verify the join/leave regex before merging this one just in case I missed something. Otherwise, I think this one is good to go and we can consider it tested now.

Thank you for working so hard to test it @DataBitz!

IceOfWraith:
Related for player count tracking: https://github.com/CubeCoders/AMP/issues/642

I'm going to merge this template since that is a minor issue and should be resolved when the sessionid is tracking users properly.

PR #147, opened 2022-06-15
Craftopia: config and server stop fixes

Greelan:


IceOfWraith (approved):
Everything looks good. No issues starting and joining with the new settings. Thanks for getting it cleaned up!

PR #150, opened 2022-06-23
Squad - Add 1007 Update Stage

IceOfWraith:
There might be an issue with Squad that it isn't grabbing 1007 which is part of the depot based on SteamDB.


Greelan:
I've found that if 1007 is part of the depot, it is included in the server download.

IceOfWraith:
Hmm... I haven't noticed this behavior. I have seen some games that include the files and still include that in the depot though.

PR #152, opened 2022-06-24
Add Vintage Story

Greelan:
Adds Vintage Story dedicated server.

Notes:
- Uses port 42420 by default.
- Configured for both Windows and Linux.
- Requires tar.exe on Windows (to extract the server download).
- Requires mono on Linux. Using Docker is recommended.
- Some settings included. More can be included but in some cases it is not currently possible to add the settings in the UI as AMP cannot handle the starting null values.

Related issues:
https://github.com/CubeCoders/AMP/issues/664
https://github.com/CubeCoders/AMP/issues/665


Greelan:
I note the Start stage included in this fails when it immediately follows the server files being unzipped/extracted. 

Unless this issue can be fixed, a pause stage would be a workaround: https://github.com/CubeCoders/AMP/issues/684

PhonicUK:
If you use a "Run an executable" stage, you can get a delay on Linux by running:

    bash -c "sleep 5"
    
Or on Windows:

    cmd.exe /C "timeout /t 1 /nobreak"

Greelan:
Well, yeah, but bit hacky? :)

PhonicUK:
I've added a Pause stage that will wait for `UpdateSourceArgs` seconds.

Greelan:
Look forwarding to testing when 2.4 is in mainline

Greelan:
With 2.4.0.2, now ready for merging

IceOfWraith (commented):
Everything seems to work well, I don't own the game to fully test though. I'm ready to give approval, but think we should discuss the picture used.

Currently, the image stretches out the new instance creation box. I provided another image from their site that doesn't do that, but then the name gets cut off in the created instance. I honestly don't care either way. But thought it should be discussed.

Greelan:
Thanks for the review. Understand the comment re the picture - I realise it stretches out the instance creation screen but it looks fine when the instance is actually created. 

I can look around for something potentially better but then again...

IceOfWraith:
I just quickly looked and didn't see anything readily available. I'll submit an approval as is, if you find something better we can figure it out later.

IceOfWraith (approved):

Greelan:
I can't find a better image that works, so will just leave as is.

PR #154, opened 2022-06-27
Mordhau - Add Arid Map

IceOfWraith:
The new Arid map was released 6/23. This adds the map to the dropdown list.


Greelan:
Looks fine to me. Although you might need to rebase the branch/PR as it is showing a bunch of older diffs.

PR #155, opened 2022-06-28
Add Empyrion Galactic Survival

IceOfWraith:
Reference: https://github.com/CubeCoders/AMPTemplates/pull/49

All changes requested by @Greelan have been implemented, but there are two things to note.
-I kept the working directory the same as the launcher lives in this folder, but calls the dedicated server executable one deeper.
-There are two ports that have unknown uses. I've looked around and have gotten mixed responses on if they're actually needed or what they do, but the telnet port is always 4 higher than the main game port in each case, so I have to assume there's a reason behind it.


Greelan (approved):
Looks fine to me, all comments addressed.

PR #156, opened 2022-06-29
Stationeers: Various updates for new dedicated server build

Greelan:
Template updates to accommodate the new dedicated server build.

Currently the main issue is that the new server console does not work within AMP. The server uses some form of console capture to enable input (using `Dear IMGUI` I believe), which causes errors when run under AMP. This results in console input and output stopping, with both Windows and Linux. I am trying to figure out how it operates to see whether it is possible to get it working within AMP.

The new dedicated server is also under active development, and settings are changing. The template is configured for the currently applicable use of `-loadlatest`, but that will change soon.


Greelan:
It appears that TCAdmin can handle the new Stationeers console via their ConsoleWrapper application: https://help.tcadmin.com/ConsoleWrapper

Apparently TCAdmin can also work with the Wreckfest console, possibly by this means too. 

Wonder if there are any learnings here for AMP's console management?

Greelan:
With 2.4.0.2, start/wait/stop stages have now been re-added, and the updates are now as far advanced as they can be. Console input is still an issue, but that is dependent on the server devs. Admin commands can now be configured for in-game

IceOfWraith:
Do I really have to test this one? Ugh...

IceOfWraith (changes requested):
Everything seems to be in order in terms of how well Stationeers works. The only thing I note is that this game does have join/leave messages. (Couldn't get the chat menu to even open though)

Join: `> 00:43:35: Client IceOfWraith (76561197977084314) is ready!`
Leave: `> 00:49:58: Client disconnected: 1 | IceOfWraith connectTime: 101.8s, ClientId: 76561197977084314`

Greelan:
Nice. I can include

Greelan:
I've now updated the regex, fixed a couple of other issues I noted, and renamed the old template.

IceOfWraith:
This is probably your level of OCD, but can you please change it to `(Legacy)` to match other templates? :p

Greelan:
The difference is that the other templates are legacy *templates* (but the same game server), whereas in this case I am highlighting that the template is for the legacy *game server*

IceOfWraith:
That seems like the same thing in this case. A template that is no longer used. It's like how you named The Isle (Legacy) too.

IceOfWraith:
I'm not pushing it any further than that. Lol I don't care enough. It just looks odd in the list.

Greelan:
With The Isle (Legacy), that is actually how the devs refer to the server to distinguish it from EVRIMA. 

I still also see this as different to the retired official modules. But ultimately the question is: whose OCD wins?

Greelan:
OK, I will change it

IceOfWraith:
Hehe

IceOfWraith:
I'll test the new regex tomorrow. It all looks good at first glance though.

IceOfWraith (commented):
The legacy seems fine.

The new server leave regex is flaky at best. I'm trying to figure out why AMP is struggling with it since your syntax is all good. It sometimes isn't registering player joins even though regex101 shows it's accurate. And never logs player leave.

Greelan:
Possibly there are rogue extra spaces in the log lines?

I discovered that with the app.ready regex.

Check the AMP logs or actual server logs for clues (AMP's console normalises for stuff like that).

IceOfWraith:
You're exactly right, but it's in the middle.. I accented where the issue was below between the username and connectTime. The join may be fine though. I haven't seen it do that again after changing this.

`> 21:14:31: Client disconnected: 1 | IceOfWraith SPACE SPACE TAB connectTime: 50.0s, ClientId: 76561197977084314`

Greelan:
Dumb, dumb, dumb.

So try `\s+` between the two bits? Like I did for app.ready

IceOfWraith:
Works like a charm! Go ahead and push the change and I'll submit an approval.

Greelan:
Done

IceOfWraith (approved):
Smooth like butter. Good work as always!

PR #165, opened 2022-07-19
Add Post Scriptum

southnode-zac:
Added Post Scriptum (effectively a clone of Squad Dedicated Server)


IceOfWraith:
I'll review this later this week. Thanks for the submission!

IceOfWraith:
> I'll review this later this week. Thanks for the submission!

As you can tell, I'm a convulsive liar. I'm looking at this now.

IceOfWraith:
Argg... My Linux setup is borked. I'll need to fix it this week and get back to you. 

One thing I note right away is to remove `Dedicated Server` from the name so it matches the others in the list.

southnode-zac (commented):
Changed name from Post Scriptum Dedicated Server to Post Scriptum

southnode-zac:
@IceOfWraith done - however I do note that Squad contains the following (which is the kvp I duplicated here as they are practically the same game, just with a different AppID)

App.DisplayName=Squad Dedicated Server

IceOfWraith:
Thank you! That one should be changed as well. I'm finally able to do testing again since the ghost left my network. Lol I'll give this a try now!

IceOfWraith:
Looks like you corrected one, but missed `Meta.DisplayName=Post Scriptum Dedicated Server`. After that this seems good to merge. Thanks for the submission!

IceOfWraith (approved):
Great first submission!

PR #168, opened 2022-07-22
Add Veloren

Greelan:
Adds Veloren dedicated server.

Note:
- Edit: Key settings are configured via template merge (given that AMP cannot work directly with RON format files). The configuration supplies a template config that the user needs to edit for any manual changes needed that are not in the UI, which is then merged to form the actual config used by the server. RON allows comments, which makes it easy to explain this in the files.
- The update stages are a little clunky given that the FetchURL update stage does not allow the target file name to be specified, and accordingly the file downloaded is missing the .zip extension and is not recognised as a ZIP by AMP for extraction purposes. This requires the file to copied first to a file with a .zip extension, and then extracted. The result is that the original download is left behind.
- The template is affected by this [bug](https://github.com/CubeCoders/AMP/issues/642), relating to tracking user joins and leaves by `sessionid`.


IceOfWraith:
One suggestion for the YAML file. I used the same trick on Empyrion since JSON is a subset of YAML. Run the contents through a site like this, then include the converted with your config if it works. AMP can then write to it since it's JSON and the game can probably write to it since it's also still a type of YAML. (That's the idea anyways)

https://jsonformatter.org/yaml-to-json

Greelan:
I misspoke, the format is RON. Corrected above.

IceOfWraith:
Lol I should have known that, but slept since we talked about it. Is this ready for testing?

Greelan:
Yep. ThisIslandEarth on Discord has put it through its paces too.

The lack of ability to specify the ports in advance bothers me. I asked in their discord whether there were any undisclosed command line params for setting ports, but they confirmed there weren’t and that they had no plans to add any params (which I find weird). The only way that AMP could in theory work with the file is via a template merge (both because of the RON format and the way the port is specified in the file along with the IP), but then that becomes a problem in terms of user control of the file, which will be needed for certain settings. More fundamentally, in some cases the AMP config manifest just won’t be able to accommodate the values required, which defeats the purpose of a template merge. Plus also a maintenance issue.

IceOfWraith:
What about putting the template in the same directory as the server's config file and naming it something obvious like "...EDITME"?

Greelan:
I don’t follow. The server generates its own config file on first start, and the contents differ depending for example whether IPv6 is enabled on the system. The devs are also adding settings over time - even “The Book” (their online manual) doesn’t have all the settings that are currently in the config file that is generated by the server

IceOfWraith:
I'm suggesting to include a template config that if the user wants to make changes to the config file manually, they make it to the template file instead of the one the server is using. AMP would overwrite the server config file with the changes made in the template file.

I just don't see any other way than a template config to have this one be merged. I understand, it'd take more maintenance over time, but that's true of all game servers that have a setting file included. I wasn't aware the contents of the file changes. Is it possible to include a majority of the settings? Then if any minor ones are missing or get added by the devs, the user can easily add them into the template file themselves?

Greelan:
I see what you mean now, but it is still messy, say with the IPv6 binding.

And many settings use the RON Option type. Which means that a setting can either be `None` or `Some(123)` or `Some(“text”)`. `None` is used when you want it to be null or empty, such as an empty string, and `Some` is where you want an actual value. The config manifest can’t handle that.

IceOfWraith:
That's so terribly messy. Who designed this document format? Don't answer that, it's rhetorical. But either way, I hate whoever it is.

What if it only handles the main settings like ports and server name? Put the IPv6 binding as `[{{$IPv6Binding}}]` and default value of `::`. Then the user can change that through the GUI. But the ports and IPv4 binding is set by AMP? Just trying to think of ideas that would make this one work without being a total manual process for the users.

Greelan:
Template merge is all or nothing - you can’t have only some settings configured via the manifest. And the IP and port are not single settings, they are combined settings `IP:port`

IceOfWraith:
I don't follow now. You can leave the other settings as they come. Don't have AMP manage those and it will copy it verbatim. If the user edits the template then the server config gets replaced with their manual edits as well. You can also specify two variables in one line (Can't you?)

Greelan:
Right, you want them to edit the AMP template rather than the server config directly. I can see that becoming confusing. Plus if an assumption is made that IPv6 is configured, but it is not, it will probably bork the server. But if you leave it to the user to add, they have to figure out how independently (not currently documented)

Greelan:
@IceOfWraith, you've shamed/inspired me into adding the template merge for key settings :)

IceOfWraith:
I'm glad you were shamed into doing it. Heh I'll take a look tonight and try out the remaining configs you have waiting!

Greelan:
I note the Start/Wait/Stop update stages that I would otherwise have included in this have been removed because the Start stage fails when it immediately follows the Set Executable stage. 

Unless this issue can be fixed, a pause stage would be a workaround: https://github.com/CubeCoders/AMP/issues/684

IceOfWraith (dismissed):
*7 years later*
Looks good.

Greelan:
Thanks @IceOfWraith. I've updated the GH path for the config file download, so please review again :)

IceOfWraith (approved):
Better do it now or it'll be another 4 weeks. xD

PR #171, opened 2022-07-23
The Isle (EVRIMA): add settings

Greelan:
- Adds new QueuePort settings
- Adds new Discord setting
- QueryPort retained but does not appear to be used anymore, so endpoint URI removed. Cannot get definitive feedback on The Isle discord as to whether used in any circumstances, so keeping as a configured port for the time being


IceOfWraith (commented):
Nothing seems off here. Good work! (as always)

IceOfWraith (approved):
Helps if I choose "Approve" instead of "Comment"...

Greelan:
I'm seeing some other issues now with the devs' latest server updates which are quite strange. I will continue to look into that and update this PR once I've got more clarity on that.

Greelan:
@IceOfWraith, now made some further tweaks:

- Fixed the QueueEnabled key as despite the devs saying they would change it from a certain version, they haven't
- The Discord setting causes issues for the server starting at least on Windows, so I am removing it
- LinuxBufferedConsole gives more console output on Linux, but messes with the metrics because AMP monitors `socat` instead. WindowsBufferedConsole stops all console output on Windows, so I have kept the AppCompatibility at None

IceOfWraith (approved):
You're the best.

PR #172, opened 2022-07-30
Pavlov VR: update user join/leave regex

Greelan:
Revised regex so that it also registers user leaves on Shack servers. Currently user leaves work on PC servers only. At the expense of getting player names in AMP console UI


IceOfWraith (approved):
Works based on testing in Discord.

PR #173, opened 2022-08-05
Update V Rising and Satisfactory Chat Regex

IceOfWraith:
V Rising and Satisfactory have no chat regex, so AMP was populating a default which caused the output to appear as chat anytime there was a `:` in the line.


Greelan (approved):
I prefer `^$` but if a space achieves the outcome, all good

PR #181, opened 2022-08-28
Broke Protocol: Add Windows support

Greelan:
Notes:
- Needs the game to be owned on Windows.
- Given SteamCMD on Windows can't download as anonymous (since the game needs to be owned, and the server is run through the game files on Windows), the template is now configured to prompt for Steam credentials. However, this means that the server now prompts for a username and password also on Linux, even though the Linux Steam update stage is only for downloading the Steam client components on that platform (not the game or the server itself). Nevertheless, `anonymous` username and password works for Linux, so the game does not need to be owned to install the Linux server.


Greelan:
Previous console output issue on Windows now fixed

IceOfWraith (approved):
That's dumb. This game dumb. Thanks for adding it though.

PR #184, opened 2022-09-03
Add tModLoader 1.4+ & Edit tModLoader (Legacy)

IceOfWraith:
Due to the major changes in how tModLoader functions, a new template has been created to address the new methods. This version will download the appropriate dotnet version and execute the tModLoader.dll. This shares the same config and metaconfig files as the original tModLoader.


IceOfWraith:
This is ready for testing. For 1.4 I've included the fixes @jasperfirecai2 created for the new journey mode.

I've added a second fix to hardcode the 1.3 version GitHub files as they use the same repo as 1.4 and this caused issues downloading.

IceOfWraith:
The download path for the config file will need updated before merging.

tmodloader14metaconfig.json
```diff
@@ -0,0 +1,8 @@
+[
+	{
+		"ConfigFile": "tModLoader14serverconfig.txt",
```
jasperfirecai2, line 3:
```suggestion
		"ConfigFile": "tmodloader14serverconfig.txt",
```

jasperfirecai2: config merging won't work if capitalisation is different

jasperfirecai2 (dismissed):
image
This message might be a candidate to put into the throwawaymessageregex. 

The server seems to work as intended on linux, cannot comment on windows. ModLoading all works. I've updated https://github.com/jasperfirecai2/amp-templates/wiki/Managing-mods-with-tmodloader which should allow users to handle the new way it works without needing any changes to the kvp.

Make sure the download path is updated as well. If you need any help with testing @Greelan send me a message

IceOfWraith:
Very nice! We should probably add what you have there to Discourse in the knowledge base and pin it in Discord in the #generic channel. In fact... I think I'll start a megathread of Generic config special requirements on Discourse and point all the channels to look there for the one source of truth.

IceOfWraith:
@Greelan It should be up to speed now. My dev branch has the latest for Legacy and 1.4 if you wouldn't mind verifying. :)

Greelan (approved):
You've taken on board feedback over discord, nothing further. Good stuff

PR #191, opened 2022-09-17
Add Blackwake

Greelan:


IceOfWraith:
Arc and I were able to test this one out. It "works" in its own special way. Lots of crazy errors about the bots spoken in another language. But otherwise, it worked. No leave regex so that's useless. It also took a couple restarts before the server actually worked. No reason behind that.

Greelan:
Thanks, repeating my comments on Discord for the record (I should have included more info in the PR but I was tired 😀 ).

I noticed as well there is no leave logs, that is why I didn’t add the join regex. No chat is logged either (I tested all three types of chat).

I did test this server with the client. The console spam and lots of errors happened on my servers too, both Windows and Linux. But the server starts and opens the ports. I don't recall that only happening after a restart - I think it just takes some time. May be wrong though. I did read somewhere about the first start being for installing various components so that is possible. Unfortunately I can't add start/wait/stop update stages because the App.Ready mode is Immediate, given there are no sensible log lines for server start regex.

I included only a few settings atm coz I don’t like wasting time on dead ends but if this is workable I can add more.

IceOfWraith:
I'd say it's workable. But also might be a waste of time for the number of uses it'll get. Lol

Greelan:
If that was the test, I'd have had a lot of my time back in the past. *Looks at Hurtworld*

Greelan:
Took about 30 mins to add everything using my spreadsheet method, so not a biggie. Some settings aren't really documented but I will include as much info as I can. 

I will just tidy it all up and do final testing.

Greelan:
All additional settings from Server.cfg added to the UI.

It does indeed appear that the first start does not open ports and seems for server setup purposes. I could add start and stop stages separated by a long pause, but I'd imagine that the server startup time will be too variable between systems for the pause to be reliable. So will just rely on users (user?) to jiggle the handle

IceOfWraith:
Completely fair in my eyes. Ready for a review?

Greelan:
Yes, thank you sir. Although would be good for @southnode-zac to also have a look, since he seems to be the only person running the server lol

I guessed at the function/descriptions of some settings, coz they don't seem to be documented, but he might have more info to refine them (eg min and max values etc)

Greelan:
Actually thinking I don't need the config download. I only did that so I could set ports on first server start, to avoid port conflicts, but as noted the server does not seem to open ports on first start. It does generate the file tho, so subsequent starts should be fine.

Greelan:
So without the config download, on the first start the server does seem to temporarily open the game port (the default) but then shuts itself down. AMP then restarts it and both game port and query port are opened. So it seems the server itself is set up to do a first run shutdown on first start, which does not happen if the config is supplied.

I tried it with two instances, one already running on the default game port and the other doing a first start, to see if the temporary port conflict creates an issue. The second server did report that it could not open a socket, but then just proceeded to shut itself down and then be restarted by AMP, this time on the right port, because by then the config file was created and the ports were set properly for the restart.

tl;dr: I will remove the config download stage

IceOfWraith:
Very nice. I'll take a look tomorrow.

IceOfWraith (approved):
˙ǝɹǝɥ pooƃ sʞool ƃuᴉɥʇʎɹǝʌƎ ¡ʞɹoʍ ʇɐǝɹפ

PR #198, opened 2022-09-27
Update Pavlov VR & Space Engineers Docker Policy

IceOfWraith:
Until we figure out why these don't work in Docker, I'm marking it NotSupported.


Greelan:
Please confirm that with that policy set, Pavlov and SE still appear in the dropdown to create instances? Previously setting the policy to NotSupported meant that they did not appear at all, even if the new instance defaults were set to not create instances in containers

IceOfWraith:
Oh it still appears. But it also let's you create them in a container. Leaving this anyways to link the issue I'm about to open. xD

IceOfWraith:
Related: https://github.com/CubeCoders/AMP/issues/736

Greelan (approved):
Approving on the basis of the responses provided, particularly that instances can still be created

PR #201, opened 2022-09-27
Update Valheim and ValheimPlus

IceOfWraith:
This adds the release stream option to allow for "default_old" that was recently added. This also enable crossplay functionality.


IceOfWraith:
This can be tested from my dev branch using `IceOfWraith/AMPTemplates:dev`.

Greelan:
The changes look functional to me, but query whether it is desirable for crossplay to be enabled by default?

This might be another case where getting IsFlagArgument to be capable of being defaulted to False would be useful. The new FlagValue does not appear to work.

IceOfWraith:
Done.

Greelan:
And that actually works? My recollection is that AMP can’t handle the empty false value and it still shows as true

IceOfWraith:
Yeah it defaults to a blank value. I think that was solved a while back with allowing null values.

IceOfWraith:
Is there anything left you'd like to see before approving?

Greelan (approved):

PR #209, opened 2022-10-05
PZ: Updates for new ports and regex

Greelan:
Deals with the new ports system introduced with 41.77


IceOfWraith (approved):
Server started fine with the new settings and everything seems in order. Good work!

PR #218, opened 2022-10-23
Add Unreal Tournament 2004

Greelan:
Adds Unreal Tournament 2004 dedicated server.

Thanks in particular to @TABLE272 for all their extensive testing and feedback.

And big thanks to @IceOfWraith for hosting the server files in a format that is workable with AMP!

Notes:

- Configured for Windows and Linux.
- Needs 2 sequential ports (game port + next port), plus an additional webserver port. AMP can't at this stage guarantee the ports are sequential.


IceOfWraith (approved):
Looks good!

PR #221, opened 2022-10-24
The Isle (Legacy and EVRIMA): general updates

Greelan:
- Improves console output
- Updates app ready regex for more appropriate log line
- Switches to new ports system (Legacy)
- Streamlines update stages


IceOfWraith:
It seems you removed Linux support, was it not working or something?

Greelan:
> It seems you removed Linux support, was it not working or something?

The Isle (Legacy) has never had Linux support - it has always been Windows only. I’ve had no success with either Wine or Proton. All I did here was remove the redundant stuff for proton in the kvp

IceOfWraith (approved):

PR #223, opened 2022-10-29
Add Night of the Dead

Greelan:
Adds Night of the Dead dedicated server


IceOfWraith (dismissed):
I was able to start/stop/update the server just fine. All settings seemed in order. I don't own the game so couldn't test connectivity, but it sounds like another user provided some testing already based on our Discord discussion.

IceOfWraith (approved):

PR #226, opened 2022-11-05
Add Impostor

Greelan:
Adds Impostor server (open source private server implementation for Among Us).

To connect to the server, client configuration is required: https://impostor.github.io/Impostor/


impostor.kvp
```diff
@@ -0,0 +1,88 @@
...
+App.TelnetLoginFormat=login {0}
+App.UpdateSources=@IncludeJson[impostorupdates.json]
+App.Compatibility=None
+App.SteamUpdateAnonymousLogin=False
+App.SteamForceLoginPrompt=False
+App.RapidStartup=False
+App.MonitorChildProcess=False
+App.MonitorChildProcessWaitMs=100
+App.MonitorChildProcessName=
+App.SupportsUniversalSleep=True
```
IceOfWraith, line 67: It doesn't seem to support sleep mode. It doesn't wake up on connection requests.

IceOfWraith (changes requested):
Looks good aside from the picture being ridiculously huge and the sleep setting. ;)

impostor.kvp
Greelan: It does on Linux. Seems the Windows bug strikes again. For that reason, and because the instance crashes if the Stop button is clicked while the server is sleeping, I've disabled Sleep.

impostor.kvp
```diff
@@ -0,0 +1,88 @@
+Meta.DisplayName=Impostor
+Meta.Description=Impostor Dedicated Server
+Meta.OS=Windows, Linux
+Meta.Arch=x86_64
+Meta.Author=Greelan
+Meta.URL=https://github.com/Impostor/Impostor
+Meta.DisplayImageSource=url:https://github.com/Impostor/Impostor/raw/master/docs/images/logo_458.png
```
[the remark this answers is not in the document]
Greelan: Done.

impostor.kvp
IceOfWraith: *Facepalm emoji*

PR #229, opened 2022-11-15
Add Call of Duty 4: Modern Warfare

Greelan:
Adds Call of Duty 4: Modern Warfare dedicated server.

Notes:
- Sadly, no console input or output with AMP.
- Authentication Port and Master Port also do not appear to listen.


IceOfWraith:
Unfortunately, I can't test this until the SteamCMD login issues are fixed.

call-of-duty4mwmetaconfig.json
```diff
@@ -0,0 +1,16 @@
...
+        "ConfigFormat":"set {0} \"{1}\"",
+        "ConfigFormatRegex":"^set (?<key>.+?) \"?(?<value>.*?)\"?$"
+    },
+    {
+        "ConfigFile":"main/server.cfg",
+        "AutoMap":true,
+        "ConfigType":"kvp",
+        "ConfigFormat":"sets {0} \"{1}\"",
+        "ConfigFormatRegex":"^sets (?<key>.+?) \"?(?<value>.*?)\"?$"
+    }
```
IceOfWraith, line 15: Why do you have the same thing twice? Or am I just blind... which is very possible.

call-of-duty4mwconfig.json
```diff
@@ -0,0 +1,478 @@
...
+        "DisplayName":"Server Name",
+        "Category":"Server Settings",
+        "Description":"Sets the name of the server. Colour codes can be used: ^0=Black, ^1=Red, ^2=Green, ^3=Yellow, ^4=Blue, ^5=Cyan, ^6=Pink, ^7=White, ^8=Grey, ^9=Grey/Map Default",
+        "Keywords":"server,name",
+        "FieldName":"sv_hostname",
+        "InputType":"text",
+        "IsFlagArgument":false,
+        "ParamFieldName":"sv_hostname",
+        "IncludeInCommandLine":false,
+        "DefaultValue":"^1AMP Powered Call of Duty 4: Modern Warfare Server",
```
IceOfWraith, line 94: The server name gets cut off after `AMP Powered Call of Du`. I'd suggest shortening it and noting the character limit in the description.

IceOfWraith (changes requested):

call-of-duty4mwmetaconfig.json
Greelan: You are indeed blind. One has the `set` prefix, the other `sets`.

IceOfWraith: Lmao it still took me a couple times to see the difference. Should probably get my eyes checked again.

call-of-duty4mwconfig.json
Greelan: Is that just a steam browser issue or does it also happen in-game?

IceOfWraith: This is an in-game limitation. Every server in the list is that many characters or less.

Greelan: OK. So 22 chars, or 24 including the colour formatting?

IceOfWraith: Sorry for the slow response. 22 characters + 2 color formatting or 22 characters without color formatting. 22 is all that will be displayed in the browser.

Greelan: Updated.

IceOfWraith:
Very good. Do the update source change and I'm a happy camper.

IceOfWraith (approved):

PR #231, opened 2022-11-16
Create Mod dir on base game update.

gkuba:
When the game is updated it will create the Mods dir if it is not present.


IceOfWraith (approved):
Thanks for knocking this out so quickly! :)

PR #238, opened 2022-11-23
Add Frozen Flame

Greelan:
Adds Frozen Flame dedicated server

The documentation on the settings is a bit ... sparse, so some of the descriptions might need improvement. Not clear either whether the last setting is actually a number or a boolean.


Provant15:
Strange that they use False/True in other areas, but knowing the game I believe it would be a boolean.

Here is also an example of how it's used on G-Portal.
image

Greelan:
Makes sense. I will get it updated. Any other feedback welcome too. AMP server logs for user joins, leaves and chats would be helpful to complete the regex

IceOfWraith:
I don't own the game to test if the server is saving properly on close. But I noticed two things about it.
1. On Linux it gives the following error after a lot of lines about exiting: `Exiting abnormally (error code: 130)`
2. On Windows it gives no feedback when stopping.

Do you know if that works properly?

Provant15:
> 

I can confirm that it works properly, and the game is playable. It does (appear to) save and pick up where the character left off when leaving/joining or restarting the server.

For some reason the console throws up a bunch of fairly useless stuff.

IceOfWraith:
Perfect. Just what I expected but wanted to verify. It's not the first game server to behave that way unfortunately.

IceOfWraith (changes requested):
Everything seems in order. I could start the server fine. I don't own it to test, but the other user here appears to have tested already. Change the update source and this is ready to go. Great work on another one!

IceOfWraith (approved):

PR #239, opened 2022-11-25
Updated stable settings and config options.

gkuba:
Updated Undead Legacy stable settings file and config options to match the new stable build. 

Experimental got pushed to Stable. Nothing else looks like it changed on Undead Legacy's side.


IceOfWraith (approved):
Very nice! Thank you

PR #243, opened 2022-11-28
Update longvinter.kvp

xLeviNx:
Just support multihome (IP choice) for multiple instances of servers

### DIFF
--- a/longvinter.kvp
+++ b/longvinter.kvp
@@ -24,7 +24,7 @@ App.ExecutableLinux=1639880/Longvinter/Binaries/Linux/LongvinterServer-Linux-Shi
 App.WorkingDir=1639880
 App.LinuxCommandLineArgs=
 App.WindowsCommandLineArgs=-FullStdOutLogOutput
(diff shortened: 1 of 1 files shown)


Greelan:
I don't think this is the right approach to this. ApplicationIPBinding is not otherwise used by the template so it isn't necessary to use it for this purpose, particularly as this approach will require GenericModule.kvp to be edited anyway. And changing ApplicationIPBinding doesn't change the displayed endpoint. 

I will add a setting that allows the MultiHome IP to be added via the UI, like a number of other templates.

IceOfWraith:
Or perhaps, we use this as a cause to push an issue request to ask Mike to make it editable in generic?

Greelan:
Useful addition, but again won't make any practical difference in this case. 

Better yet, there should be a dropdown of available interface IPs, like in the official modules. Plus an empty field to avoid the flag being used at all (needed eg for Ark).

Greelan:
For Longvinter, the devs have changed how the default Game.ini is handled and also some of the contents. So will add those as well

Greelan:
See https://github.com/CubeCoders/AMPTemplates/pull/246

xLeviNx:
Closed ; refer to #246

PR #262, opened 2022-12-27
Updated Killing Floor 2 config. Starting map/Mutators 

DeteCT0R:
I did remove the default starting map list for a text box so you can put any starting map what u want.(Default KF2 or Workshop map)
Also added a Mutator field even u can add mutators in KF2 config file or KF2 intergrated webadmin but some mutators need to be loaded from command line else its gonna crash the server.


killing-floor-2config.json
```diff
@@ -69,6 +30,18 @@
             "KFGameContent.KFGameInfo_Objective": "Objective"
         }
     },
+    {
+        "DisplayName": "Mutators",
```
IceOfWraith, line 34: Are there other things besides mutators that this could be used for? As in, should this just be a "Customer Arguments" input or are mutators all a person would ever add?

killing-floor-2config.json
```diff
@@ -2,53 +2,14 @@
     {
         "DisplayName": "Starting Map",
```
IceOfWraith, line 3: Rather than remove all the options, I'd recommend you take a look at Empyrion for how the Scenario and Custom Scenario inputs work. It should apply here as well.
https://github.com/CubeCoders/AMPTemplates/blob/main/empyrion-galactic-survivalconfig.json

IceOfWraith (changes requested):
Thanks for looking into this one! I've provided a couple notes for feedback, please feel free to respond with the changes, or reasons why I'm wrong (which is pretty often). :)

killing-floor-2config.json
DeteCT0R: Hmm. Technically yeah you can use the field to override any paramter after maxplayer(The fact you can do that from server also) so yeah "customer arguments" would be the best for it. For executing "bad" stuff im not sure the field is positioned in the middle of the command line and the server is already started. I will dig deeper when i have more time how i can do the best way and i will come back.
Kinda rare when people modding KF2 server so not really important part but its always good to have.

killing-floor-2config.json
DeteCT0R: I will check it.

DeteCT0R:
Okay i changed the maplist now its including the default maps and a custom option. For mutators i tested multiple method which is the best and safe. For now by default its including an empty ?Mutator= parameter in the command line its not looking best but its stop people for override any other parameter in the command line and only accept Mutator.

IceOfWraith:
I'm working on testing this now. You'll need to replace manifest.json with the original one from this repo before this can be merged though.

killing-floor-2config.json
```diff
@@ -2,7 +2,7 @@
...
         "Category": "Server Config",
-        "Description": "Set the server's starting map. Survival compatible maps denoted with \"*\".",
+        "Description": "Set the server's starting map or a custome map by choose \"Custom Map\"",
```
IceOfWraith, line 5: The grammar is incorrect here.

killing-floor-2config.json
```diff
@@ -11,44 +11,62 @@
         "IncludeInCommandLine": false,
         "DefaultValue": "KF-Bioticslab",
         "EnumValues": {
-            "KF-Airship": "Airship",
```
IceOfWraith, line 14: I'm not sure why you decided the map names shouldn't be user friendly?

killing-floor-2.kvp
```diff
@@ -22,7 +22,7 @@ App.ExecutableLinux=232130/Binaries/Win64/KFGameSteamServer.bin.x86_64
 App.WorkingDir=232130
 App.LinuxCommandLineArgs=
 App.WindowsCommandLineArgs=
-App.CommandLineArgs={{Map}}?Game={{Game}}?maxplayers={{$MaxUsers}} -Port={{$ApplicationPort1}} -QueryPort={{$ApplicationPort2}} -WebAdminPort={{$RemoteAdminPort}} -Multihome={{$ApplicationIPBinding}} {{$FormattedArgs}}
+App.CommandLineArgs={{Map}}?Game={{Game}}?Mutator="{{ServerMutators}}"?maxplayers={{$MaxUsers}} -Port={{$ApplicationPort1}} -QueryPort={{$ApplicationPort2}} -WebAdminPort={{$RemoteAdminPort}} -Multihome={{$ApplicationIPBinding}} {{$FormattedArgs}}
```
IceOfWraith, line 25: I haven't tested myself yet, does this still load properly and keep the max players if you leave the Mutator blank?

IceOfWraith (changes requested):
I've provided some feedback if you can please look it over and provide responses/fixes where needed. Sorry for taking so long to get a response in on this one.

PR #264, opened 2022-12-29
Update GatekeeperV2

IceOfWraith:
This removes the requirement to use a special Docker image since the python3 tagged ampbase image includes pip now. This also adds the additional stuff into a script that runs with the update stages instead of being handled by the before mentioned image. These are very quick installs that take only a couple of seconds.


Greelan (approved):
Fixed a typo (and some minor formatting issues)

PR #265, opened 2022-12-29
Stationeers: update container policy

Greelan:


IceOfWraith (approved):
I assume there's nothing Mike can fix in the Docker images since it isn't available at all on Debian based on your reason given?

Greelan:
Version 2.36 will be available in Debian 12 (bookworm): https://packages.debian.org/search?searchon=sourcenames&keywords=glibc

It could probably be installed manually on Debian 11 but that could cause other issues.

Greelan:
I do note however that even with the container policy changed, I can still create a dockerized Stationeers server on Linux (which ofc doesn’t work). The policy reason text also does not show. Sigh…

IceOfWraith:
There can be some funkiness with that where AMP thinks you're running in or not in docker until ADS is bounced. Perhaps that's the issue?

Greelan:
OK, an ADS restart fixed it. Thanks

PR #269, opened 2023-01-02
Update MinValue for all player counts

IceOfWraith:
This is to address https://github.com/CubeCoders/AMP/issues/803


Greelan (approved):
Aside from being triggered by where you placed the field ( :) ), this looks fine. Glad you picked up on the -1 min for Eco.

PR #270, opened 2023-01-04
Updated ecoconfig.json

southnode:
Changed MaxConnections to DefaultSlots - maxconnections was removed from the Network.eco setting, and replaced with MaxConnections


IceOfWraith (approved):
Very nice. Thanks for the addition!

PR #272, opened 2023-01-07
SteamCMD client app ID updates

Greelan:


IceOfWraith (approved):

IceOfWraith:
I created an instance using Assetto Corsa using the 2.4.1.2 release and didn't run into issues, so this seems to be backwards compatible.

PR #281, opened 2023-01-21
Update Valheim+

IceOfWraith:
This updates Valheim+ to remove the spaces in the ini file to match the original file. This also clarifies the description for Valheim Plus Enable.


Greelan (commented):
I think this config setting and the SteamCMD and Updates setting should note that the server needs to be updated after enabling Install V+. Otherwise it reads that it is enough just to enable the setting. 

You probably don't need now ConfigFormat and ConfigFormatRegex.

Greelan (approved):

PR #288, opened 2023-02-10
Project Zomboid: setting tweaks

Greelan:


IceOfWraith (approved):
You fixed it before I could submit a review. Haha

PR #292, opened 2023-02-25
Eco update

southnode:
Change ecoconfig.json line 73 from text to Textarea - allows for multi-line detaileddescription fields. Tested and works (including testing with the colour system and it works still)


Greelan:
You’ve got other guff in there that needs to be cleaned up.

Create the branch based off the CC repo, not your own.

Issue #285, opened 2023-01-27
Svencoop - Missing maxplayers/ Stuck at starting


DeteCT0R:
Svencoop config missing the `+maxplayers ` from the command line. (And not in the config) without that the server start with default maxplayer which is 8.
Also the server stuck at "Application State: Starting" even u see the `Connection to Steam servers successful.` message (which is should be the "ready" message.)
After that u cant start the server because the same port already using.
`Sock_Open: WARNING: port: 27001 bind: Address already in use`

PR #293, opened 2023-03-02
Sven Co-Op: Add maxplayers

UVClay:
Commit adds maxplayers as a configuration option in the Server Settings area.  Going to work on fixing the server not stopping/restarting properly after this but I wanted to get this up now as I saw there was an open issue for it as well.


UVClay:
Solves #285

Greelan:
I've added some tweaks. Does the argument really need to be in both command line and config file?

UVClay:
Nope, just the command is fine.  Sorry, I haven't been using AMP all that long and didn't realize it was duplicating the setting in to server.cfg.

Greelan:
Seems like it is not the only setting being done that way. Probably avoids any confusion. Shouldn't present an issue.

sven-co-opconfig.json
```diff
@@ -12,6 +12,21 @@
...
         "EnumValues": {}
     },
+    {
+        "DisplayName": "Max Players",
+        "Category": "Server Settings",
+        "Description": "maxplayers",
+        "Keywords": "maxplayers,limit",
+        "FieldName": "$MaxUsers",
+        "InputType": "number",
+        "MinValue": "1",
```
IceOfWraith, line 22: It seems the MinValue should be 2, and can you please add a MaxValue of 32?

IceOfWraith (changes requested):
Thank you for the submission! Sorry it took so long to get back to you. (and everybody lately) I've given feedback on a minor change. Please update or provide a response on why I'm wrong and this one can be merged.

IceOfWraith:
> I've added some tweaks. Does the argument really need to be in both command line and config file?

This is how you asked me to do it previously in case someone edits the config file and the startup arg doesn't take priority. Probably came from that.

sven-co-opconfig.json
Greelan: Done it for you, lazy

IceOfWraith: I was wanting feedback since I've never played this game. Just going by a guide. And as you know, mods add changes to games that could change the max value for things like this. Wanted to avoid that scenario.

Greelan: Well, that’s fair.

IceOfWraith: I'm not finding any proof that there's more than 32 allowed even with mods. So lets go ahead and close this one. Thanks for updating it!

IceOfWraith (approved):

sven-co-opconfig.json
xLeviNx: Yeah max is 32 regardless of mods. Engine limit

IceOfWraith: Awesome. Thanks for confirming!

PR #301, opened 2023-03-18
DayZ: FPS and filepatching settings changes

Greelan:


IceOfWraith (approved):
Using variables within other settings still impresses me. Great work!

PR #306, opened 2023-03-21
Various: Wine updates for 32 bit servers

Greelan:
Updates the templates for Assetto Corsa, Carrier Command 2 and Call of Duty 4: Modern Warfare so that Xvfb is no longer required to run the servers on Linux.


IceOfWraith (approved):
It's very pretty. Thank you for this!

PR #307, opened 2023-03-21
Various: updates to RCON settings

Greelan:
Introduces changes to various templates that use the server's RCON, to make it clearer how the user can configure the RCON password in AMP (and deprecates previous settings that no longer work).


IceOfWraith (approved):
Very nice getting all this cleaned up! You even caught the 7d2d changes I've been meaning to make but keep forgetting on the keywords. The only thing I wonder, is it a good idea to change the FieldName's to something else just to be safe?

Greelan:
Please clarify your query?

IceOfWraith:
> Please clarify your query?

Should the fields that refer to $RemoteAdminPassword be changed to something else to avoid possible future issues?

Greelan:
That's deliberate. In those cases the relevant setting is the conduit for the setting being applied to the relevant config. So while it can't be used to actually set the password, it is the bridge between the kvp and the server config.

IceOfWraith:
Obviously. I overlooked that simple fact. Lol

IceOfWraith (approved):

PR #309, opened 2023-03-22
Update Mordhau

southnode:
Add in SteamWorkshopDownloadLocation (even though the mods are on mod.io), add in custom map in the enum for starting map & mode, and add additional box to specify custom map
Remove RCON Password setting from Server Config - replace with info tab
Set Admin Password input to password type instead of random for UX
Set default admin password to a random new guid as placeholder for user.


mordhau.kvp
```diff
@@ -20,6 +20,7 @@ App.RootDir=./mordhau/
 App.BaseDirectory=./mordhau/629800/
 App.ExecutableWin=629800\Mordhau\Binaries\Win64\MordhauServer-Win64-Shipping.exe
 App.ExecutableLinux=629800/Mordhau/Binaries/Linux/MordhauServer-Linux-Shipping
+App.SteamWorkshopDownloadLocation={{$FullBaseDir}}Mordhau/Plugins
```
IceOfWraith, line 23: Since this doesn't use the Steam Workshop section, this should be blanked out or removed entirely. If this is kept blanked out, the version needs bumped on `Meta.MinAMPVersion`.

IceOfWraith (changes requested):
Thanks for the submission! One change was requested. Otherwise, it works as intended. Was able to load a mod map after fumbling through how to do mods on Mordhau. x)

mordhau.kvp
southnode: @IceOfWraith is there a way to hide the Steam Workshop section completely considering it does not use them?

and when you say blanked out, you mean
`App.SteamWorkshopDownloadLocation=` without anything after it?

Whichever way this works, I will make the amendment and bump the version

IceOfWraith: There's no way to hide it. I've asked Mike for that multiple times with no memorable response.. It's very misleading and clutter. Just another UX thing that should be improved and more thought out.

The way you changed it there would be fine. Then bump the version to 2.4.3.0 and it should be good to go.

southnode: Done and done

IceOfWraith (approved):
Very nice!

PR #312, opened 2023-03-23
Add Foundry Virtual Tabletop

Greelan:
Adds Foundry Virtual Tabletop dedicated server

Note:
- Both Linux and Windows are supported - the template is configured to download its own Node.js, and so is able to run on each platform regardless of whether Node.js, or the correct version, is otherwise installed on the system. The UI allows selection by the user of the desired version (must be at least v14.x)
- The user is required to supply a (timed) download URL from the Foundry VTT login portal in order to download/update the server
- Minimal settings are configured given that the server has a web interface for configuring them, and in many cases AMP can't handle them anyway (due to starting `null` values)


IceOfWraith (approved):
As discussed, I don't own so only a desktop review. Looks good though.

IceOfWraith (approved):
Everything seems okay, as much as I can test without owning it.

IceOfWraith (approved):

Greelan:
Thanks. It's been tested by a couple of other people in Discord and no issues.

I've updated the extra steps URI to point to Discourse.

Greelan:
After the next AMP release I will add ARM support.

PR #314, opened 2023-03-26
Add Skyrim Together Generic Template

RFrost619:


Greelan:
The template doesn't actually install the server? I don't think it can be merged in this state.

IceOfWraith:
Yeah... There was discussion on Discord around this one but I don't think the convo was ever finished.

IceOfWraith:
Here's the conversation. It's mixed in with another game so couldn't remember where we left off. (Next time you should include these details in the PR. Lol) The issue is the files are only available through Nexus Mods site and you have to sign in to get the link. It honestly looks to be simpler to upload the files to AMP in this case since the alternative is to copy the link into an input but that also might be confusing for users.

https://discord.com/channels/266012086423912458/671396391804731405/1088265006749077614

Greelan:
I don't think it is confusing. It's the same issue for Foundry VTT - see my PR.

RFrost619:
You would have to copy the link from your browsers download tray. 
Example link: https://supporter-files.nexus-cdn.com/1704/69993/Skyrim%20Together%20Reborn-69993-1-4-1-1679265299.zip?md5=tkywp_miw_12A56EbtzaLw&expires=1679866440&user_id=[USERID]&rip={YOUR.IP.ADDR.HERE}
I did attempt to add the zip download but it's not completing for some reason. Updated my fork/dev branch.

IceOfWraith:
Hmm... I  don't see an issue with your config in the dev branch right away. Just to be sure, it's not downloading it fully versus just not extracting it?

Greelan:
Closing this in favour of https://github.com/CubeCoders/AMPTemplates/pull/454. But the bulk of the work is the OP's, so great job!

Greelan:
BTW, @RFrost619 , do you know if the server requires the "Address Library for SKSE Plugins" installed, or is that only client side?

Also, can you share the log lines that you based your user join and leave regex on?

PR #318, opened 2023-03-28
Conan Exiles: various updates

Greelan:
Switch to new port system, add Wine env vars, change AdminPassword setting to be user configurable, other tweaks


IceOfWraith (dismissed):
Good stuff! Keep them coming!

IceOfWraith (approved):

Greelan:
Now updated to set the AdminPassword via ini command line param. As a result the config download is no longer needed (keeping the config in the repo for legacy installations)

Greelan:
Switching back from using new ports system, given the Refresh Configuration logic may not be able to handle it

Greelan:
Reverted ports switch, now it has the new ports system again

conan-exilesconfig.json
```diff
@@ -161,7 +158,7 @@
         "IncludeInCommandLine":false,
         "DefaultValue":"0",
         "EnumValues":{
-            "0":"Europe",
+            "0":"Europe (default)",
```
IceOfWraith, line 161: You knew I'd ask this. Is there a benefit to calling out a default when it is different for all users?

IceOfWraith (approved):
All looks good other than the one question

conan-exilesconfig.json
Greelan: Fixed

IceOfWraith (approved):

PR #322, opened 2023-03-29
Various: Wine related updates 

Greelan:
Makes various Wine related changes to the templates that use Wine:

- Adds env vars to ensure the prefix is specific to each instance
- Adds an update stage to initialise/update Wine for better operation
- Adds Wine 8 docker support to Icarus, The Isle (Legacy) and Night Of The Dead to enable them to be run on Linux. Also switched Empyrion: Galactic Survival to Wine 8 docker to address changes to the server (plus a metaconfig fix)


IceOfWraith (approved):
Phenomenal work! It's clear how much time and effort you put into this.

PR #329, opened 2023-04-07
No One Survived

Partiesplayin:
So I'm not sure how this works exactly but I would like this game reviewed for compatibility template for native use with AMP

https://store.steampowered.com/app/1963370/No_One_Survived/


Greelan:
If you want to request a game being added, create a request under https://github.com/CubeCoders/AMPTemplates/issues, supplying all the info asked for.

PR #335, opened 2023-04-09
Terraria - Allow changing steamcmd beta branch

jasperfirecai2:
Also Uses more readable update sources


IceOfWraith (approved):
I made a few changes to match other templates. This looks good though. Thank you for the submission!

This will be merged when the AMP release happens tomorrow. There's currently an issue with updating existing templates that this would cause issues for users.

PR #341, opened 2023-04-19
Add Mindustry

Greelan:
Adds Mindustry dedicated server.

Notes:
- Installs its own Java 17 to run the server.
- All relevant settings configured.
- Allows user to specify release version of the server to install.
- Includes ARM support. 
- No SteamCMD mod/map support given that AMP does not behave sanely in terms of requiring login.


IceOfWraith (approved):
Desktop review looks good!

PR #343, opened 2023-04-23
Frozen Flame: updates for server changes

Greelan:
The perils of creating templates for early access servers....


IceOfWraith (approved):
Everything appears to be in order. Nice work as always!

PR #351, opened 2023-05-03
Project 5 sightseer

darkace1998:
Tested on windows and linux

no problems found


Greelan (approved):
With tweaks added, seems to work fine

PR #359, opened 2023-05-10
Risk Of Rain 2: Add mod functionality, tweak settings and metaconfig

Greelan:


darkace1998:
i tested your template server installes normal and starts without problems only thing is that we get a lobby bug on joining the dedicated server gitting this error on winodows linux and from lan and wan connections

Greelan:
What do you mean by "lobby bug"?

darkace1998:
Lobby bug is join can join the server but you cant select any characters and skills.

its just a blank lobby without any context

darkace1998:
@Greelan i fixed the problem you can only join a dedicated server if cross play is disabled in the client game settings

darkace1998:
after looking at the template mode i found that the docker restrictions(CPU,RAM limits) are not working.

and the Port allocation is broken the server only looks at port 27015 even if u deploy multiple templates.

IceOfWraith:
It sounds like you're having unrelated issues to this template. I'd tale to the Discord in the tech-support section.

Greelan:
On the second issue, you may not have the latest templatel

IceOfWraith (approved):
All worked find here. Nice work getting the mods working!

PR #364, opened 2023-05-14
Pavlov VR: Fix docker support

Greelan:
All along it was AMP...

Also updates ports, tweaks settings


gOOvER:
still `bEnabled` not in config. its for visibility of the server in serverlist

IceOfWraith:
> still `bEnabled` not in config. its for visibility of the server in serverlist

Where are you seeing this? It's been in the config all along.

IceOfWraith:
I realized you're stating that the setting isn't exposed in the UI. This game only lets you join from the in-game server browser AFAIK. There's no direct connect option. So exposing that would just confuse people and make them unable to join. It's in the ini file as true if someone wanted to change it though.

IceOfWraith (approved):
Awesome job figuring this one out. This is the most bizarre one yet.

PR #365, opened 2023-05-15
Arma3 and DayZ: simplify mod handling

Greelan:
Use AMP's now inbuilt mod symlinking

### DIFF
--- a/arma3.kvp
+++ b/arma3.kvp
@@ -22,6 +22,7 @@ Meta.NoCommercialUsage=False
 App.DisplayName=Arma 3
 App.RootDir=./arma3/
 App.BaseDirectory=./arma3/233780/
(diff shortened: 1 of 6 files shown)


IceOfWraith (approved):
Seems good to me. Is there a reason for the script versus just putting it directly into the update stage?

IceOfWraith:
I should have asked "What's the reason?" because there's always a reason with you. Lol

Greelan:
I tried it directly - that is my strong preference - but with all the escaping required due to it being included in json, it didn't work. Maybe I didn't translate it properly, but I didn't have hours to devote to figuring it out.

IceOfWraith:
Completely understand that. Lol

PR #368, opened 2023-05-16
Update assetto-corsa-comp.kvp

hanunes:
Updated UserJoinRegex and UserLeaveRegex to reflect the true nature of the behaviour.


Greelan:
I can't recall - is the sessionid and userid always numbers only?

And is the time of "5s" in the new regex always the same?

hanunes:
sessionid is always a number (incremental one). userid is the steamid, which is always in the same format, starts with S and has only numbers ahead of it.

5s has to do with the ignorePrematureDisconnects setting, so I'm 99% sure it is, if not, can always make it a regex

Greelan (approved):
Thanks for the contribution

PR #378, opened 2023-05-21
Update OpenTTD

IceOfWraith:
Due to the latest update, the save file must be removed, the game started and an autosave created, then the save game location put back in before starting the server again.

### DIFF
--- a/openttdconfig.json
+++ b/openttdconfig.json
@@ -115,7 +115,7 @@
     {
         "DisplayName":"Save Name",
         "Category":"OpenTTD Server Settings",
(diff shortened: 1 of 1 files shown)


IceOfWraith:
I didn't think about the description. I was hesitant to remove the default as then if people do start it twice it will make a new world. I think at least.

Greelan:
But with the default there it won't start tho right?

IceOfWraith:
It would fail. And I was kind of banking on that. But either way I don't think is a good method. It sucks they don't make it more automated. But it's also a free game, so how much complaining can be done?

See how you feel about the updated version.

Greelan (approved):

PR #387, opened 2023-06-14
Pavlov VR: update stage updates

Greelan:
The update to v29 of the PC server recently has changed the server binary name. But the shack version has not changed. This update accommodates both.

<s>(Separately there is an issue with the v29 PC server where the shared library `libc++.so` needs to be redirected to a more modern version, `libc++.so.1`. But that can't be handled in the template given the files are owned by root.)</s> This issue is now addressed.


Greelan:
The most recent commit caters for the different shared library location in Docker.

IceOfWraith (approved):
Wow! You deserve a trophy for being the best damn template maker here.

PR #390, opened 2023-06-15
Pavlov VR: Failsafe paths

Greelan:
Cater for spaces (I should know better)


IceOfWraith (approved):
Oops. Easy to forgot.

PR #394, opened 2023-06-17
Add Staxel

Greelan:
It seems a dead project, but oh well


gOOvER:
staxel dont run in wine, so it would be only a Windows-only game

IceOfWraith (approved):
Looks good!

PR #412, opened 2023-07-01
Update Randompassword Inputs

IceOfWraith:


Greelan (commented):
Suggest `{{newguid()}}` for those settings that otherwise supply a default value

IceOfWraith:
Good call. Got them updated now.

Greelan (commented):
Don't add it to SE or The Forest

IceOfWraith:
I placed it on all so they'd have a default secure password in case the user never puts one in.

Greelan:
Thought you were big on dev defaults? :)

For many servers, blank password = not enabled.

Anyway, for SE at least it seems the UUIDv4 format doesn't work anyway, so what's the point?

IceOfWraith:
I said I *prefer* to keep dev defaults. But they aren't always good with security. Lol

Anyways, I've made the changes if you want to review.

Greelan (approved):

PR #421, opened 2023-07-15
Update path-of-titans.kvp

southnode:
Update docker image to deb12 version post successful testing


IceOfWraith:
I made a request to have all images migrated to Debian 12 which should resolve this. Thanks for the contribution!

PR #427, opened 2023-07-23
Sons of the Forest: Update Regex

winglessraven:
Update regex for app ready, user join, and user leave


IceOfWraith (approved):
Good stuff! I just made a couple changes for the "Please restart" portion and to accommodate if the client number gets to the double digits.

PR #435, opened 2023-07-29
Update conan-exiles.kvp

IceOfWraith:
Update to include non-Steam client mod downloading port.


Greelan (commented):
Should it be "Game and Mods Port"?

Also, would it be worth hardcoding the offset (hopefully an ini command line arg works) so that users can't change and mess up the port setup?

IceOfWraith:
You're right. Even after you corrected what I put I still read it wrong. Lol

Can you elaborate what you mean for the offset?

Greelan:
From the wiki:

`Default value: 0 (offset added to the game port)
Config file: Engine.ini, section: “[/Script/ConanSandbox.SystemSettings]”, entry: “ModFileDownload.DownloadPortOffset=2”`

So if the ini command line arg works it would be:

`ini:Engine:[/Script/ConanSandbox.SystemSettings]:ModFileDownload.DownloadPortOffset`

With a hardcoded value of 0 set in a hidden config manifest setting

Greelan:
Or just add directly to the command line args (like I did for The Isle):

`-ini:Engine:[/Script/ConanSandbox.SystemSettings]:ModFileDownload.DownloadPortOffset=0`

IceOfWraith:
Ahh okay. I follow. Yeah let me do that. The ini command line doesn't work last we tried but I can put it in the config manifest.

Greelan:
It worked some of the time for AdminPassword if I recall... Lol

But the hidden setting to overwrite it if someone adds it to Engine.ini will also address it, given the default is what we want

IceOfWraith:
Well, that won't be necessary. Because it's already the second setting in the config manifest. x)

Greelan:
Lmao, you pre-emptied yourself

Greelan (approved):

PR #440, opened 2023-08-09
Astroneer: fix port argument, other tidy-ups

Greelan:


IceOfWraith (approved):
Looking good!

CostcoFanboy:
Is there a reason that Astroneer is not available for Linux? Is wine no-go? Ptero made it run so I'm just curious.
No pressure to do it or anything, I'm just curious as I thought it was simply abandoned.

Greelan:
It requires encryption to be disabled on both server and client to work under Wine

CostcoFanboy:
I checked on Ptero and all the client would need to do is disable encryption from the client.

Why not leave it server-side by default True, then make a setting to be able to turn it to False and explain the steps to disable it client-sided and repercussions. While it sounds awful to disable security measures, I think it's better than flat out not being able to play the game.

Just 2 cents of course, thank you again for your work.

IceOfWraith:
I think we're all pretty much in agreement that we don't want to implement that "fix". It's a terrible solution and personally I wish the devs hadn't offered it. It's a cop-out for a real fix on their side.

CostcoFanboy:
> I think we're all pretty much in agreement that we don't want to implement that "fix". It's a terrible solution and personally I wish the devs hadn't offered it. It's a cop-out for a real fix on their side.

The devs will never give enough of a poop to fix it. It's been more than 3 years. I don't think it's worth holding out in the hope that they do something personally.

> Plus it will result in continual support issues. People don't read.
I imagine that happens with like, every game already. Pinned comments and keep making people read those🤷🏻

Greelan:
You're welcome to try the testing template in my dev branch to see that if achieves the purpose. `Greelan/AMPTemplates:dev`. It will show under "Greelan / Astroneer [Linux Testing]"

I've implemented the encryption disabling in the simplest way possible tho this method does always work for Unity servers

Greelan:
Now merged

CostcoFanboy:
You guys move faster than I thought 😓 sorry I have like 3-4 Github accounts because of my work so I often have trouble getting back to people on my personal one.

PR #462, opened 2023-09-09
Add Operation: Harsh Doorstop

Greelan:


IceOfWraith (commented):
I couldn't see the server in-game to connect. It does show up in my Steam favorites though. It doesn't let me check for this game on the Steam public list. Got any ideas? Otherwise, I don't see any issues here.

Greelan:
According to the guide, all three ports (including rcon port) need to be forwarded for the server to show in the in-game browser: https://docs.google.com/document/d/1JFxwBBBJoiBqQ9bgAFAGCLd_lpu8FXD6XxaKcMiZgg8/edit

Seems odd to require the rcon port, but there it is

IceOfWraith:
That's a royal pain. I'm assuming there's no feasible way to make the port a regular port instead of the RemoteAdminPort so it's opened by AMP?

Greelan:
Huh? I think you are confusing the SFTP port

IceOfWraith:
I thought RCON wasn't opened by default for some reason. I don't ever use it. Lol

IceOfWraith:
No idea what's up then. Mine are auto-forwarded  and everything should be fine. But if you've already verified the connection works, I don't mind approving. Just note there may be a strange issue with this one.

Greelan:
Haven't verified anything with this one. Someone asked for it, I made the template, they have not given any feedback

IceOfWraith:
It's free on Steam if you want to try it.

IceOfWraith:
Ahhhhh hold the phone. I may have found the issue...

Greelan:
I know. It only runs on Windows tho

IceOfWraith:
It looks like this is another case of "ARK" where they hide non-password protected servers by default and I have to tick the box to show them. Waiting to see if it shows up now though.

IceOfWraith:
Of course, I had internet issues right after sending that. But it's not showing still. I'll leave it running until tomorrow to see if it's a timing issue.

IceOfWraith:
I'm really not sure about this one. I've tried following some YouTube guides and matching up things, but they had false info in them as well (like put your public IP as the binding IP)... I just can't get it to show in game and therefore can't connect. I tried adding `-NOSTEAM` as one video suggested, no go. I tried adding a MapCycle.cfg and it reads it but no change. I'm listed as an admin, it has a password... Here's a pic of the in-game browser and filters in case I'm missing something dumb.

image

Greelan:
@IceOfWraith based on your testing, I don't think there is anything more needed on this one? As noted, logs are not useful for user joins and leaves

IceOfWraith (approved):

PR #464, opened 2023-09-21
Squad Changes

southnode:
Change verbiage and defaults on configurations
Change regexes for join and leave
Disable writable console


Greelan:
Has the regex been tested with multiple players connected? I suspect that given there is no identifier for individual players, AMP will treat one player leaving as providing a match for all players - so the player count will drop immediately to 0

Greelan:
Either way, this PR has conflicts that will prevent merge anyway. It needs to be rebased

southnode (commented):
@IceOfWraith re-reviewed this. Those two files are good to go.

PR #473, opened 2023-10-01
Add Counter-Strike 2

Greelan:
The current srcds module for CS:GO could be switched to the `csgo_demo_viewer` branch for those who still want a CS:GO server (tho who knows how long Valve will keep that around).


counter-strike2config.json
```diff
@@ -0,0 +1,212 @@
...
+        "FieldName":"sv_tags",
+        "InputType":"text",
+        "ParamFieldName":"sv_tags",
+        "IncludeInCommandLine":true,
+        "EnumValues":{}
+    },
+    {
+        "DisplayName":"LAN Server",
+        "Category":"CS2 Server Settings",
+        "Description":"If enabled, the server will operate as a LAN server: no heartbeat, no authentication, no non-class C addresses, 9999.0 rate, etc",
```
IceOfWraith, line 28: It's unclear what 9999.0 rate means here. Is that the network speed?

counter-strike2config.json
```diff
@@ -0,0 +1,212 @@
...
+        "ParamFieldName":"sv_cheats",
+        "IncludeInCommandLine":true,
+        "DefaultValue":"false",
+        "EnumValues":{
+            "False":"false",
+            "True":"true"
+        }
+    },
+    {
+        "DisplayName":"Disable Valve Anti-Cheat",
```
IceOfWraith, line 165: Not a required change by any means, but when I read this it felt backwards as opposed to a setting that was "Enable Valve Anti-Cheat" that defaults to on.

IceOfWraith (approved):
It seems there's a bug with CS2 that you can't use a password protected dedicated server right now. It doesn't prompt for a password and gives an error that the password is bad in game. Otherwise, it runs like I'd expect! I gave some feedback that you can choose to disregard if you prefer.

counter-strike2config.json
Greelan: Given the server by default (ie without anything specified) enables VAC, I think it makes more sense to do it this way.

counter-strike2config.json
Greelan: I will remove that reference. It is a hangover from the srcds description of the sv_lan setting and is not referenced in the help text for the CS2 cvar, so can be removed.

IceOfWraith (approved):

PR #480, opened 2023-10-11
Add Call of Duty: Modern Warfare 2 (2009)

Greelan:


IceOfWraith (approved):
I don't own this one, but from a desktop review it all looks good.

PR #486, opened 2023-10-16
Update euro-truck-simulator-2.kvp to include windows executable

tomtenberge:
insert missing windows executable


Greelan:
It's not "missing" as such. It was omitted by @IceOfWraith because he found that the server did not run properly under NETWORK_SERVICE, which is the account that AMP generally runs instances under on Windows.

This is explained in the docker policy reason in the template: "The game server does not run properly as NETWORK SERVICE, so it must be run within Docker on Windows."

Have you tested your change?

IceOfWraith:
Moreover, if it was tested, do you have AMP running as a service?

Greelan:
That's the issue. On most AMP installations on Windows, AMP runs as a service, so that it starts on boot. So the native server won't work for most Windows installations

PR #488, opened 2023-10-20
Add The Front

Greelan:


IceOfWraith (approved):
Wow! My eyes are spasming after looking over this one. I can only imagine how you felt. Lol it looks good though. I only found one typo which should be fixed now. Nice work!

IceOfWraith (approved):

PR #492, opened 2023-10-23
Update Call of Duty MW3 to add year

ssfdre38:
Added Year to Call of Duty MW3 Title due to new Call of Duty MW3 coming out on Nov 10th.


Greelan (approved):
Consistency changes made

PR #494, opened 2023-10-27
Add CryoFall

Greelan:


Greelan:
Windows support removed because it has no console output or input (and OS_CLOSE does not work) and therefore the server cannot be shut down properly

IceOfWraith (dismissed):
It seems to work if you don't have conflicting ports. :)

Desktop review only, but seems fine.

IceOfWraith (approved):

PR #507, opened 2023-11-05
Add CryoFall UserJoin and UserLeave RegEx

winglessraven:
Add CryoFall UserJoin and UserLeave RegEx.

UserChat will not work because it spans multiple lines, eg...

```
5.11.23 09:02:08.144 [IMP] [SCRIPT] ChatId=13 From="winglessraven": 
test
Character "winglessraven" (Id=5307)
```


Greelan (approved):
Thanks!

PR #526, opened 2023-11-16
Redbot Discord Bot to AMP

ssfdre38:
This is to add Redbot Discord Bot to amp. I know that there was some talk in the development channel about this and this is just to see if Mike would want to have bots to be added to AMP or not.


IceOfWraith:
As discussed in Discord it was decided universal templates like the NodeJS one should be used instead of individual templates for each bot. We can discuss on a case-by-case basis if there's some kind of exception to that. A good contribution would be converting the existing GatekeeperV2 template to allow any Python app to work with it similar to the NodeJS one. As well, the NodeJS template could use some updates to find a way to allow running npm or npx as the startup commands versus node.

Greelan:
I started yesterday developing a generic Python bot runner template and will have it completed shortly.

PR #529, opened 2023-11-17
Update CoreKeeper UserJoin and UserLeave Regex

winglessraven:
Update CoreKeeper UserJoin and UserLeave Regex to match current output.

Console output for join:
`[userid:76561197972547399] player winglessraven connected`

Console output for leave:
`Disconnected from userid:76561197972547399 with reason App_Min` (for disconnect)
`Disconnected from userid:76561197972547399 with reason AppException_Max` (for alt+f4)


Greelan:
Looking good. Note that there could be an issue with AMP failing to register leaves properly as a result of the joins matching on two vars but the leave only of one (have encountered that bug in other contexts)

Greelan (approved):

winglessraven:
> Looking good. Note that there could be an issue with AMP failing to register leaves properly as a result of the joins matching on two vars but the leave only of one (have encountered that bug in other contexts)

Thanks, I did wonder that but worked OK in testing

PR #553, opened 2023-12-13
Update squad-dedicated-server.kvp

southnode:
Update to add EOS compatibility port

### DIFF
--- a/squad-dedicated-server.kvp
+++ b/squad-dedicated-server.kvp
@@ -1,10 +1,10 @@
 App.AdminMethod=SourceRCON
 App.ApplicationPort1=7787
 App.ApplicationPort2=27165
-App.ApplicationPort3=0
+App.ApplicationPort3=15010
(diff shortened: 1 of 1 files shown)


IceOfWraith (approved):

Greelan:
Had this been tested on an existing instance? Given AMP won't add the port to existing instances, the command line arg will not be populated properly. Curious if that breaks the server

southnode:
> Had this been tested on an existing instance? Given AMP won't add the port to existing instances, the command line arg will not be populated properly. Curious if that breaks the server

Yeah unfortunately existing instances flat out don't work as-is, and will need to be redeployed. Developers practically introduced EOS backend switch overnight. Caused all sorts of mischief.

Greelan:
Still, AMP should handle adding the ports, so you wouldn't need to totally redeploy. Such a missing link

southnode:
Agreed. If ports are added with a new config, AMP should integrate them.

PR #554, opened 2023-12-14
BeamMP: Update Source Fix

TABLE272:
It appears the last few BeamMP releases are using different file names, leading to the server files failing to download.

I've removed "linux" and "archlinux" which seem to no longer be in use.
Changed the new defuault to the "ubuntu" build, which is Ubuntu 22.04 and made this the default based on the description mentioning this build was recommended for AMP's Debian 12 container. Updated the description to match.
Debian 11 build remains available under "debian"

Based on the one user I was helping test this, updating this may require existing users updating their config to work around the list bug, changign the dropdown then changing it back, due to removal of the old value. If there's a better way to handle that go right ahead :)

### DIFF
--- a/beammpconfig.json
+++ b/beammpconfig.json
@@ -2,7 +2,7 @@
     {
         "DisplayName":"Linux Server Version",
         "Category":"BeamMP - Server Settings",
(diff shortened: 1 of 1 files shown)


Greelan (approved):

Greelan:
Thanks. There are means to try to cater for instances that have the older setting, but it's a lot of additional work for small gain. People will just have to toggle

PR #555, opened 2023-12-15
Change Post Scriptum to Squad 44

southnode:
Developer changed the name of the game. No mass underlying changes to the game code, folders remain the same name for now.


Greelan:
Not sure why you wouldn't just change the two DisplayNames and the ServerName default value, and leave the rest so that existing instances aren't left stranded?

IceOfWraith (approved):

PR #570, opened 2023-12-25
Add Smalland: Survive the Wilds

Greelan:
Currently configured to run the Windows server under Wine on Linux, as the devs botched the Linux server release. Hopefully in the new year they will switch to Steam download for both platforms


IceOfWraith (approved):
Desktop review looks good. What a messy release.

PR #579, opened 2024-01-03
V Rising: add BepInEx option to Linux

Greelan:


IceOfWraith (approved):
Everything looks good here.

PR #580, opened 2024-01-03
Various: update binding settings

Greelan:
Caters for 2.4.7


IceOfWraith (approved):
You are an absolute beast! Nothing sticks out after reviewing these. Thank you so much for getting it updated!

PR #588, opened 2024-01-10
V-Rising Scheduled Wipes

Psychloor:
Add support for scheduled wipes for v-rising games.
parameter names can be found in: https://github.com/StunlockStudios/vrising-dedicated-server-instructions
which I used

### DIFF
--- a/v-risingconfig.json
+++ b/v-risingconfig.json
@@ -387,7 +387,7 @@
       "False": "false"
     }
   },
-    {
+  {
     "DisplayName": "Scheduled Wipe Interval",
     "Category": "V Rising - Host Settings",
     "Description": "Interval in days between each wipe. 0 = Disabled",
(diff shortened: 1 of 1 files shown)


Greelan:
Thanks for this. Are those two settings in the server's default ServerHostSettings.json file?

Also, no need to have two Any settings for the wipe days - just pick one. I suggest AnyDay.

Greelan:
If they are not the file, the settings won't work

Psychloor:
> If they are not in the file, the settings won't work

Worked when I added them to the serverhostsettings file in savedata folder

Psychloor:
But alright got it

Greelan:
> > If they are not in the file, the settings won't work
> 
> Worked when I added them to the serverhostsettings file in savedata folder

Yeah, because then they are ... in the file. Lol

PR #591, opened 2024-01-15
Add tmlsavedirectory to args

jasperfirecai2:


jasperfirecai2:
in my testing it worked. Try removing the hidden config value so that it only uses the command line arg, it will still save to the configured value.

Greelan:
You mentioned something on discord about the server complaining or erroring, so I wondered

Greelan (approved):

PR #599, opened 2024-01-19
Add Palworld

TylorMayfield:


Greelan:
My dev branch already has a much more developed template

PR #604, opened 2024-01-22
Add Mount & Blade II: Bannerlord

Greelan:
Minimal settings configured given that there is a separate web admin panel run by the server, and settings differ depending on the server mode being run and so are better left to the panel/config files.


IceOfWraith (approved):
Looks good. Definitely minimal.

PR #605, opened 2024-01-23
Update palworld.kvp

rezor92:
I don't know if I'm wrong but I think there are some "-" missing :) If that is not correct just close the PR.


Greelan:
I've just followed exactly what the devs have included in their docs. It works fine as is. The dashes might work also but no need to change.

PR #628, opened 2024-02-11
Valheim: Add save interval parameter

emidah:
Hi! This PR adds support for the Valheim world save interval command line parameter (-saveinterval \<seconds\>). It was added to the game [sometime in 2022](https://store.steampowered.com/news/app/892970/view/3376031059808269643), but I saw AMP doesn't have it yet!

From the "Valheim Dedicated Server Manual":

image

This is my first contribution here, so forgive me if I got something wrong (should it be left empty by default instead?)

### DIFF
--- a/valheimconfig.json
+++ b/valheimconfig.json
@@ -267,6 +267,20 @@
     "Special": "listfile:./Valheim/896660/Saves/bannedlist.txt"
   },
   {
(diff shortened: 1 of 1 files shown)


Greelan (approved):
Thanks!

PR #659, opened 2024-03-09
Stormworks: updates

Greelan:
The devs appear to have deprecated the server app id (no longer being updated), and the server files bundled now with the game files do not work on Linux using either Wine or Xvfb/Wine (whether Wine stable or Wine staging). Congrats to the devs.


IceOfWraith (approved):

IceOfWraith:
So disappointing.

PR #661, opened 2024-03-10
Add Dead Matter

winglessraven:
[Dead Matter](https://store.steampowered.com/app/2184150/Dead_Matter/) got off to a very rocky start when launched into Early Access. However, with the latest update the game is finally moving in the right direction.

This template adds support for Windows and Linux (via Proton) - worth noting that nobody else in their Discord has managed to get the server working on Linux yet, so AMP is currently the only Linux supported source.

Only issue of note is that user leave/ join regex does not work yet (shows in logs but not in the console) - hopefully future development of the game server will fix this.

> 
>**About the game:**
>
>The year is 2025. A mysterious virus, which turns its victims into ravenous inhuman husks, plunges the world into total chaos.
>
>Several weeks have passed since the infected have overrun the province of Alberta. Canadian federal agency NERA's evacuation attempts failed, leaving you and countless other survivors trapped in the Kananaskis area, now overrun with tens of thousands of infected pouring over from the cities.
>
>Your supplies are running low, and you’re not the only one starving and desperate.
>
>Who will you be, now that nothing is left?
>
>**What is Dead Matter?**
>
>Dead Matter is an open-world, multiplayer zombie survival game project in Early Access.
>
>We are a small but passionate team that believes in working closely with our community to refine and improve our game. As part of this collaborative approach, we offer access to a Nightly testing branch where all of Dead Matter's latest changes and newest content can be experienced before stable patches roll out.
>
>Patches and updates developed through feedback from the Nightly testing branch are integrated into Dead Matter's Release branch. This process ensures that Dead Matter will continue to evolve with our community's direct input.


Greelan:
Tried `-stdout` or `-stdout -FullStdOutLogOutput` instead of `-log` to get console output?

winglessraven:
I have not, didn't know about those - will give it a go. Thanks!

winglessraven:
Last update stage is a weird one - without it the server just errors and will not start. Found that running Proton once with PROTON_LOG=1 seems to do something to give it a kick. Running Proton on subsequent runs are then fine.

The two game ports are apparently what's needed, I think the second one is a query port - will change it.

`-stdout -FullStdOutLogOutput` seems to work! I'll also change that.

Greelan:
> Last update stage is a weird one - without it the server just errors and will not start. Found that running Proton once with PROTON_LOG=1 seems to do something to give it a kick. Running Proton on subsequent runs are then fine.
> 
But how does the server stop when run in that update stage? Just crash?

winglessraven:
It doesn't start

Greelan:
I should add that `runinprefix` is not my preferred way of running Proton, but `proton run` doesn't give console output (likely because it starts a steam.exe child process) - and using socat (LinuxBufferedConsole) gives a NRE (which I am hoping Mike can fix).

Greelan:
> It doesn't start

That is super weird.

winglessraven:
> > It doesn't start
> 
> That is super weird.

You're telling me. Spent hours yesterday trying to debug.

It wouldn't run in AMP so tried outside of AMP and it would crash, then I added `PROTON_LOG=1` to try and see what was going on. It gives some error about no tracked_files, then by removing `PROTON_LOG=1` it works 🤷🏻‍♂️

Greelan:
`tracked_files` is created in `compatdata` when using `proton run`

winglessraven:
> `tracked_files` is created in `compatdata` when using `proton run`

Is it also created when using `runinprefix`?

Greelan:
No

winglessraven:
Is there a better way to create the tracked_files that I should change it to? Assuming `proton run` would need an extra step of stopping the application?

winglessraven:
Hmm, `-stdout -FullStdOutLogOutput` worked on Windows but on Linux it seems nothing is now showing in the console other than: 
```
fsync: up and running.
CAppInfoCacheReadFromDiskThread took 0 milliseconds to initialize
Setting breakpad minidump AppID = 2184150
```

PR #662, opened 2024-03-12
Add Dead Matter

winglessraven:
[Dead Matter](https://store.steampowered.com/app/2184150/Dead_Matter/) got off to a very rocky start when launched into Early Access. However, with the latest update the game is finally moving in the right direction.

This template adds support for Windows and Linux (via Wine).

If an Out of Memory error occurs in Wine then running the following command as root fixes the issue:

```
bash -c 'echo "vm.max_map_count = 16777216" > /etc/sysctl.d/20-max_map_count.conf && sysctl --system'
```

> 
>**About the game:**
>
>The year is 2025. A mysterious virus, which turns its victims into ravenous inhuman husks, plunges the world into total chaos.
>
>Several weeks have passed since the infected have overrun the province of Alberta. Canadian federal agency NERA's evacuation attempts failed, leaving you and countless other survivors trapped in the Kananaskis area, now overrun with tens of thousands of infected pouring over from the cities.
>
>Your supplies are running low, and you’re not the only one starving and desperate.
>
>Who will you be, now that nothing is left?
>
>**What is Dead Matter?**
>
>Dead Matter is an open-world, multiplayer zombie survival game project in Early Access.
>
>We are a small but passionate team that believes in working closely with our community to refine and improve our game. As part of this collaborative approach, we offer access to a Nightly testing branch where all of Dead Matter's latest changes and newest content can be experienced before stable patches roll out.
>
>Patches and updates developed through feedback from the Nightly testing branch are integrated into Dead Matter's Release branch. This process ensures that Dead Matter will continue to evolve with our community's direct input.


Greelan:
I've committed a few tidy-ups. I've also got some other questions that I will put in a review.

deadmatterconfig.json
```diff
@@ -0,0 +1,128 @@
...
+        "FieldName":"SteamServerName",
+        "InputType":"text",
+        "IsFlagArgument":false,
+        "ParamFieldName":"SteamServerName",
+        "IncludeInCommandLine":true,
+        "DefaultValue":"AMP Powered Server",
+        "EnumValues":{}
+    },
+    {
+        "DisplayName":"Lobby Header",
```
Greelan, line 16: Why isn't this setting using the same FieldName. Isn't it just the server name? Usually SteamServerName is used to display the server name in the Steam browser, which sometimes the normal server name does not do

deadmatterconfig.json
```diff
@@ -0,0 +1,128 @@
...
+        "DisplayName":"Enable PvP",
+        "Category":"Dead Matter Server Settings",
+        "Description":"If set, PvP will be enabled",
+        "Keywords":"pve,pvp",
+        "FieldName":"PVP",
+        "InputType":"checkbox",
+        "IsFlagArgument":false,
+        "ParamFieldName":"/Script/DeadMatter.DMGameSession.PVP",
+        "IncludeInCommandLine":false,
+        "DefaultValue":"true",
```
Greelan, line 67: I switched this to match the supplied Game.ini

deadmatterconfig.json
```diff
@@ -0,0 +1,128 @@
...
+        "Category":"Dead Matter Server Settings",
+        "Description":"Sets additional command arguments to start the server with that are not otherwise set by AMP",
+        "Keywords":"switches,command,line,arguments",
+        "FieldName":"ServerSwitches",
+        "InputType":"text",
+        "IsFlagArgument":false,
+        "ParamFieldName":"ServerSwitches",
+        "IncludeInCommandLine":false,
+        "DefaultValue":"",
+        "Placeholder":"-hardcore -heartbeat",
```
Greelan, line 84: Why not add settings for these known command line args?

deadmatterconfig.json
```diff
@@ -0,0 +1,128 @@
...
+        "EnumValues":{}
+    },
+    {
+        "DisplayName":"Admin",
```
Greelan, line 88: This Admin setting, and the Super Admin setting, don't work as desired. As I understand it, the Game.ini can include multiple Admin and SuperAdmin entries. However, these settings will overwrite all of them to the same value.

deadmattergame.ini
```diff
@@ -0,0 +1,7 @@
+[/Script/DeadMatter.DMGameSession]
+ServerName=AMP Powered Server
+MOTD=This server is powered by AMP
+PVP=true
+bIsHardcore=false
```
Greelan, line 5: Why is this included here, but not in the config manifest? How is this setting different to the `-hardcore` argument?

deadmatterupdates.json
```diff
@@ -0,0 +1,57 @@
...
+        "UpdateSourcePlatform":"All",
+        "UpdateSource":"CreateDirectory",
+        "UpdateSourceArgs":"{{$FullBaseDir}}DeadMatter/Saved/Config/WindowsServer",
+        "OverwriteExistingFiles":false
+    },
+    {
+        "UpdateStageName":"Game.ini Config Download",
+        "UpdateSourcePlatform":"All",
+        "UpdateSource":"FetchURL",
+        "UpdateSourceData":"https://raw.githubusercontent.com/winglessraven/AMPTemplates/deadmatter-wine/deadmattergame.ini",
```
Greelan, line 29: This will need to be repointed to the CC repo before merge.

deadmatterupdates.json
```diff
@@ -0,0 +1,57 @@
...
+        "UpdateSourcePlatform":"Linux",
+        "UpdateSource":"Executable",
+        "UpdateSourceData":"/bin/bash",
+        "UpdateSourceArgs":"-c 'WINEPREFIX=\"{{$FullRootDir}}.wine\" WINEARCH=win64 /usr/bin/wineboot --init --update'"
+    },
+    {
+        "UpdateStageName":"Winescript Download",
+        "UpdateSourcePlatform":"Linux",
+        "UpdateSource":"FetchURL",
+        "UpdateSourceData":"https://raw.githubusercontent.com/winglessraven/AMPTemplates/deadmatter-wine/deadmatterwinescript.sh",
```
Greelan, line 45: This will need to be repointed to the CC repo before merge.

deadmatterupdates.json
```diff
@@ -0,0 +1,57 @@
...
+        "UpdateSourceArgs":"winescript.sh",
+        "UpdateSourceTarget":"{{$FullInstanceDir}}",
+        "OverwriteExistingFiles":true
+    },
+    {
+        "UpdateStageName":"Install Windows Components",
+        "UpdateSourcePlatform":"Linux",
+        "UpdateSource":"Executable",
+        "UpdateSourceData":"/bin/bash",
+        "UpdateSourceArgs":"./winescript.sh"
```
Greelan, line 55: I assume you have established that these winetricks packages are actually required?

Greelan (changes requested):

deadmatterconfig.json
winglessraven: The ServerName in the Game.ini displays in the lobby (but not the server list) when you join a server. SteamServerName in the previous parameter only sets the name in the server browser, and not the lobby. The two are not linked.

Greelan: I get that they display in different places, but wouldn't they normally be the same? As you have put in as default values? In which case have them both set by the same FieldName, and one of the settings is hidden

deadmatterconfig.json
winglessraven: The format required is different and not -{0}="{1}" like the others. Is it possible to have different methods? e.g. hardcore mode is just `-hardcore`

Greelan: Of course. See the Ark template for examples. `IsFlagArgument` doesn't work properly, so you just include the FieldName (eg `{{hardcore}}` in the command line args directly the kvp

deadmatterconfig.json
Greelan: Yes, you can't do multiple in AMP (a long requested feature). You will have to leave users to include the lines manually - which is better than them only being able to include one Admin or one SuperAdmin

Greelan: In some other templates for situations like these, I have included an "info box" in the UI to tell the user how to specify the relevant setting manually

deadmatterconfig.json
winglessraven: Ah ok, that makes sense. Like this?
```json
{
        "DisplayName":"Lobby Header",
        "Category":"Dead Matter Server Settings",
        "Description":"Sets the server title displayed in the lobby",
        "Keywords":"server,name,lobby,servername",
        "FieldName":"SteamServerName",
        "Hidden":true,
        "InputType":"text",
        "IsFlagArgument":false,
        "ParamFieldName":"/Script/DeadMatter.DMGameSession.ServerName",
        "IncludeInCommandLine":false,
        "EnumValues":{}
    }
```

Greelan: Yup, but remove the keywords, otherwise the hidden setting will be displayed in search, which is nonsensical because the user can't see it

deadmatterupdates.json
winglessraven: I added them from info I was given - will try a test without them

deadmatterconfig.json
winglessraven: Updated in latest commit

deadmatterconfig.json
winglessraven: Changed in latest commit

deadmatterconfig.json
winglessraven: Changed in latest commit

deadmattergame.ini
winglessraven: Removed from Game.ini

deadmatterupdates.json
winglessraven: Not needed - removed in latest commit

deadmatterupdates.json
winglessraven: Actually runs fine without them 🤦🏻‍♂️

Greelan:
Testing just now suggests Xvfb is not needed either.

winglessraven:
> Testing just now suggests Xvfb is not needed either.

My bad - I don't think I even tried without it, started with a copy/paste of a different template and didn't think to check that. Checking now.

Greelan:
Started OK for me (took ages tho).

Is `DM_Alberta_Persistent_WP` the only map? Are there modded maps? If other standard or modded maps, suggest adding Initial Map / Custom Map settings. See various templates for approach.

winglessraven:
> Started OK for me (took ages tho).

Yeah, all good without it.

> Is `DM_Alberta_Persistent_WP` the only map? Are there modded maps? If other standard or modded maps, suggest adding Initial Map / Custom Map settings. See various templates for approach.

It's the only map (modded or otherwise).

Greelan (approved):

winglessraven:
Thanks Greelan. Appreciate the guidance.

PR #669, opened 2024-03-16
Update mount-and-blade2config.json to fix typo

SolSoCoG:
2 times MinValue, one was supposed to be MaxValue.


IceOfWraith (approved):
Good catch!

PR #673, opened 2024-03-17
Adding exclusivejoin and NoTransferFromFiltering to Ark Survival Ascended

Nypheena:
This Adds NoTransferFromFiltering = Prevents ARK Data usage between single player and servers who do not have a cluster ID. Even with a cluster id set, it is suggested to keep this option to completely disable code path allowing it.

And Exclusivejoin = Activate a whitelist only mode on the server, allowing players join if added to the allow lists. This option uses `PlayersExclusiveJoinList.txt` and `PlayersJoinNoCheckList.txt` files. The files are located in `/ShooterGame/Binaries/<PLATFORM>/` where `<PLATFORM>` is either `Linux` or `Win64` depending on host operating system. `PlayersExclusiveJoinList.txt` is used straightforward, allowing players only whose IDs are present in the file to connect in the server.

Here's my trainwreck of a contribution, I tested it on my own and it works but I'm sure it could be cleaned up better keywords and/or description wise

### DIFF
--- a/ark-saconfig.json
+++ b/ark-saconfig.json
@@ -147,7 +147,7 @@
         "IsFlagArgument":false,
         "ParamFieldName":"PlayerWhitelist",
         "DefaultValue":"",
-        "Special":"listfile:./ark-se/2430930/ShooterGame/Binaries/Win64/PlayersExclusiveJoinList.txt",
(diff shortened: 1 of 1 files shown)


Greelan:
Thanks, this reinvents the wheel a bit from the ASE template for exclusivejoin but only half does it. I will add it properly to the upcoming ASA update. As mentioned in discord, NoTransfer doesn't seem to matter

PR #680, opened 2024-03-22
Dead Matter - Update Sever Binary

winglessraven:
Update server binary to match current version


Greelan (approved):
Usually Steam uses the query port to find the server and populate it in the browser, then the game port to connect to the game. I assume you've verified the port change in the endpoint URI when clicked

winglessraven:
> Usually Steam uses the query port to find the server and populate it in the browser, then the game port to connect to the game. I assume you've verified the port change in the endpoint URI when clicked

Yep, verified with another user (I don't normally use the endpoint links).

PR #694, opened 2024-04-03
Adding missing track configs.

solarisfire:
Config options were missing for some of the DLC circuits in Assetto Corsa Competizione.

Added:

-COTA
-Indianapolis
-Red Bull Ring
-Watkins Glen


Greelan:
Are the DLCs downloaded by default?

solarisfire:
DLC would be needed client side to join the session, but server side all tracks are included.

Greelan (approved):

PR #705, opened 2024-04-05
Added some missing settings to Ark Survival Evolved

Flon22:
Added a few missing settings to the arkSE configuration, as well as adding the configtype information to the game.ini file.


Greelan:
I will have a look at this when I get a chance. Can you confirm tho that each of these settings is in the default Game.ini that AMP supplies and aren't removed by the Ark server when it runs. AMP can't put settings into Game.ini that aren't already there

Flon22:
Thanks, no rush.
Yea I did realise that when testing. I was hoping there was some protocol for adding them in, but by the sounds of it, if it isn't there when the template is first made, it can never be there?

I've had to add two settings to the game.ini as they aren't present. There are also quite a few more that I haven't added as part of this merge but could do with being added in the future.

Greelan:
Ark is tricky, particularly SE as it has a habit of removing settings itself, at least from GUS.ini. The holy grail is AMP writing stuff in itself when it is isn't there, but we've been waiting a long time for that. 

The issue with just adding stuff into the UI when it isn't or won't be in the ini is that inevitably people will ask why the setting is not working

Greelan:
Closing because not confirmed to work on existing instances without the settings already in the file

PR #706, opened 2024-04-05
Updating "exit"-command

Hedrauta:
Got changed a while ago.
It's "shutdown" for "killing" the server nicely

Discovered via "help" in RCON/Console


IceOfWraith (approved):
Thank you!

PR #707, opened 2024-04-06
Update beammp.kvp, Change shutdown-method

Hedrauta:
String-method works fine, with "exit" as "shutdown"-string

Also, more required to do so, because there are several workers that need to be killed more nicely.

Raised timeout, because workers can take some time to quit.


IceOfWraith (approved):
Thanks! Tested good on my end.

PR #711, opened 2024-04-10
Fix BepInExPack URL for Thunderstore

fclante:
Old ThunderStore URL caused an empty zip file to be written. That caused a load of exceptions when AMP instance tried to read and extract the empty zip file.


Greelan:
Already addressed in #710

PR #728, opened 2024-04-18
Update arma-reforgerconfig.json

IceOfWraith:


Greelan (commented):
What's the first change for?

I think the field should be text for the second.

Greelan:
Fix JSON arrays and then it can be a list xD

Greelan (approved):

PR #760, opened 2024-05-04
Add Stay In Tarkov Template

dzlockhead:
Added SIT Templates. Tested and working on Linux, other users reported working on Windows. Thank you for assistance from Ice, Greelan and Daniel.


dzlockhead:
This should be ready for review.

Greelan:
@dzlockhead I've added my recommended changes. Let me know what you think.

Couple of other matters:
- seems the game port should also be TCP only?
- the capitalised text in the SPT AKI Version and SIT Version settings was a bit confusing to me, maybe it can be clarified?
- is there a reason only 2 versions of SPT AKI are hard-wired?
- similarly, is there a reason for only one hardwired SIT version?

dzlockhead:
@Greelan 

Possibly they should be TCP only but the devs did not specify on their requirements page. I figured both was the safest. 
Let me attempt to clarify. I wasn't quite sure of how to word it.
Do you mean the fetch URL? What do you mean by they are hardwired?

dzlockhead:
@Greelan 
Added some additional commits, but why delete the SIT Extra files deletion step?

Greelan:
> @Greelan Added some additional commits, but why delete the SIT Extra files deletion step?

Because DeleteAfterExtract in the SIT download stage does the job xD

Greelan:
> @Greelan
> 
> Possibly they should be TCP only but the devs did not specify on their requirements page. I figured both was the safest. Let me attempt to clarify. I wasn't quite sure of how to word it. Do you mean the fetch URL? What do you mean by they are hardwired?

Re the ports, I only see them listening on TCP.

Re the versions, I mean you only allow limited specific versions to be downloaded, even though many more are available, especially for SIT.

stay-in-tarkovconfig.json
```diff
@@ -60,7 +60,7 @@
 	{
 		"DisplayName": "SIT Version",
         "Category": "SIT Server Settings",
-		"Description": "Sets the SIT version to install. THE MAJOR VERSION MUST MATCH THE SPT VERSION FOR SPT AKI",
+		"Description": "Sets the SIT version to install. THE MAJOR VERSION MUST MATCH THE SPT VERSION FOR SPT AKI. Example: When 1.6.4 with AKI 3.8.0-2dd4d9 is selected, you must have 3.8.0 as the SPT AKI version.",
```
Greelan, line 63: The example suggests 3.8.1 can never been used, because there is only one SIT version included? So what's the point of having 3.8.1?

dzlockhead: SIT for 3.8.1 is supposed to be released soon. You are correct though. The example does suggest that.

dzlockhead:
> > @Greelan
> > Possibly they should be TCP only but the devs did not specify on their requirements page. I figured both was the safest. Let me attempt to clarify. I wasn't quite sure of how to word it. Do you mean the fetch URL? What do you mean by they are hardwired?
> 
> Re the ports, I only see them listening on TCP.
> 
> Re the versions, I mean you only allow limited specific versions to be downloaded, even though many more are available, especially for SIT.

I'm not sure how to set it up to allow previous versions. Is there a way to make FetchURL work with that? The issue is that their file hosting is not standard. I figured I'd be able to use a variable but since they moved the most recent 3.8.1 on to a different site. 
For example 3.8.1 is SPT-3.8.1-29197-d3ac83e.7z BUT they moved it to a different site. No longer on custom gitea
3.8.0 is RELEASE-SPT-3.8.0-29197-2dd4d91.7z on custom Gitea
3.7.5 is RELEASE-SPT-3.7.5-26535.zip on custom Gitea

PS as I typed this out: In testing, I MAY be able to make it work. That new custom site has links that seem to work when using other versions. Hidden, but it works.

Will also take a look at SIT to see if there's a way to make it work since it does use Github

Greelan:
No idea whether it needed to include older versions, since I have never played this game xD

SIT should be easier as you could potentially use FetchURLFromJQ and the GH API to always get the latest, and potentially older versions as specified by the user

Greelan:
BTW,what's the purpose in having the start/wait/stop update stages?

dzlockhead:
There are some config files that are generated on first time startup that
need to exist for the port configs to work right. I don't believe they are
all created when extracting.

Greelan:
One of those crappy downloads, huh?

This is a bit problematic though, as if you have two servers, one on default, presumably the first-time startup of the second server will fail due to conflicting ports? Usually in these situations the template should supply its own version of the needed configs so that the ports are set correctly from the outset.

dzlockhead:
> One of those crappy downloads, huh?
> 
> This is a bit problematic though, as if you have two servers, one on default, presumably the first-time startup of the second server will fail due to conflicting ports? Usually in these situations the template should supply its own version of the needed configs so that the ports are set correctly from the outset.

So I did test this. It still succeeds in setting up the server. It will just say the port is already bound, but when the user goes to fire up the server for the first time for actual use, it works right. My concern about supplying my own is what if the devs add fields to it later? Now I'm supplying an out of date file.

Greelan:
OK that's probably OK then. 

I suspect since SIT generates these files that any new additions would likely be added on start up. You could test this by removing lines and then run the server.

dzlockhead:
@Greelan 
I absolutely believe you are correct. Hopefully not too many support queries haha. I just tested the versioning that I investigated after you made me question just how it works and now it works even happier. Unfortunately though, SPT has released 3 new versions since I started this and SIT hasn't updated but their devs say soon! But the server does start up with SPT 3.8.3 with SIT 3.8.0 but I won't provide any support if that breaks. Not meant to work that way lol.

dzlockhead:
Added versioning for SIT. Now pulls from Github.

Greelan:
Why not go the route that is done for other templates - eg for dotnet or node or BepInEx or a range of others - that have a text field for manually entering the version number? Means the template doesn't have to be updated every time there is a new release

dzlockhead:
That's true, but I'll defer to you on this, can users be trusted to spell
it right and not make mistakes? I figured emum would cause less mistakes
than a free form field.

Greelan:
Sure but with a link and instructions it shouldn't be hard. Many templates have this. Otherwise, will you stick around to update the template every couple of days? XD

dzlockhead:
> Sure but with a link and instructions it shouldn't be hard. Many templates have this. Otherwise, will you stick around to update the template every couple of days? XD

I was planning on it, I'll try the text fields but I think that will be too confusing for users.

Greelan:
> > Sure but with a link and instructions it shouldn't be hard. Many templates have this. Otherwise, will you stick around to update the template every couple of days? XD
> 
> I was planning on it, I'll try the text fields but I think that will be too confusing for users.

Having now looked at their version list and how a user would need to find out the download links for SPT, I'm inclined to agree. I will do a final pass of the template later today and it can then be merged based on the current approach

dzlockhead:
Thanks Greelan, I had prototyped the instructions for a text field. To say
they were confusing would be an understatement. Even I found it a bit hard
to understand my own instructions. If anyone pings me as SIT or SPT have
updates, I will update as necessary, but mostly focusing on SIT. They just
released a prerelease but it's not on Github yet.

Greelan (approved):

PR #764, opened 2024-05-08
VRising - Added more game settings and interaction settings

dat-data:
With the official launch releasing in ~3 hours, I figured I could update a good chunk of the settings. I split Player Interaction Settings out into its own tab/page as it was nested settings and the game settings tab now scrolls.

Tested these locally via the configmanifest.json in my VRising instance and loaded with out failure. Settings modified in the panels (Game Settings, Interaction Settings) made the appropriate changes to the SaveGameSettings.json file.

*Note: No settings were deleted, just moved to the order they show up in the SaveGameSettings.json file.


dat-data:
@Greelan @IceOfWraith @tempusthales any chance I could get some eyes on this?  I'd like to start working on getting more settings in the next few days.

IceOfWraith:
Working on it.

IceOfWraith:
I did one small change to the regex to make it optional instead of an "or". There's at least one DisplayName that doesn't have spaces in it. And the checkbox ones need EnumValues added (see VAC for the proper way to handle those).

dat-data:
Thanks, I'll get those updated. 

I did try the checkbox option in the configmeta.json file in my local instance using booleans and it worked as expected, flipping the checkbox in the UI updates the appropriate values in ServerGameSettings.json

IceOfWraith:
Working doesn't mean right though. Doing it the other way is the proper way and prevents messed up settings in case AMP changes anything in the future, too. :)

dat-data:
Can't argue with compatibility! I'll get those updated too and keep that in mind for the next batch of settings 👍

tempusthales (approved):

IceOfWraith (approved):
Thanks for adding all of these!

PR #769, opened 2024-05-12
V Rising - Added the rest of the settings

dat-data:
- Added game time, trader, vampire, equipment, unit, and castle settings
- Added more options for Game Mode Type
- Added reference to server reset options

WarEventGameSettings were left out, I simply don't know enough about war events nor could I find any documentation to reference what the settings did.


IceOfWraith (approved):
Thank you for adding all these!

PR #775, opened 2024-05-13
Adjust ARK Ascended backupExclusions

laurencei:
Hi,

I looked at one of the generated backup `.zip` files for ARK Ascended and noticed the default template is including a number of directories that are not needed.

Firstly is the base server directory including `.cache` and `.steam` directories.

Then for the game itself, it is including the static Plugins folder, along with all crash reports and logs. Whilst these have some importance, for the purpose of restoring etc, they just take up large amounts of storage for no gain.

image
image


IceOfWraith (approved):
Thanks for this! I think the new Smart Backup system makes this less needed, but I don't think everyone knows about that yet.

PR #788, opened 2024-05-17
Add Max Transfer Rate setting to Foundry

Alacard:
By adding this setting allows you to speedup save transfer speed when logging in to server


Greelan:
Can you link patch notes or whatever that have info on this setting? Its not on the foundry website for server configuration

Greelan (approved):

PR #795, opened 2024-05-20
Update Stay in Tarkov for 1.6.5 and futureproof Github integration with new setting. 

dzlockhead:
Added 1.6.5 SIT and 1.8.3 SPT. Removed incompatible SPT versions. Added SIT Release version and made it as easy as possible to understand.


Greelan:
You need to rebase your branch to remove the merge conflicts and so we can actually see what the changes are

dzlockhead:
Okay, NOW it should be good. @Greelan

Greelan:
I find this confusing tbh. Why is SITReleaseVersion needed?

dzlockhead:
It's used to choose the release tag on github. You told me that it won't
work using just a file name and variable, which is the issue I had in
testing. It worked for 1.6.4 but when I tried 1.6.5 without the
UpdateSourceVersion, it still tried to look at release 1.6.5 for the file,
so I use that to pick the release tag.

Greelan:
If you are specifying the full file name - as you are  - then just use FetchURL. My comment was related to using GithubRelease where there was not a standard name format between releases. Its unnecessary to adopt an approach that requires the user to set two variables

dzlockhead:
Makes sense! Modifications made and should be good now. @Greelan

Greelan:
Just be aware that this will break existing instances until the user toggles the setting. The alternative is to keep the existing enum values but have separate update stages for each (based on conditions). This just might become a bit unsustainable as new versions are added in the future.

dzlockhead:
> Just be aware that this will break existing instances until the user toggles the setting. The alternative is to keep the existing enum values but have separate update stages for each (based on conditions). This just might become a bit unsustainable as new versions are added in the future.

Do you think there'd be a better alternative?

Greelan:
I suggest taking the hit now so that it is sorted for the future

Or get the SIT devs to name their files more helpfully xD

dzlockhead:
We can only hope. Let's go for it and take the hit now haha.

Greelan:
You've got the SPT AKI Version setting in there twice

dzlockhead:
I am not sure how that occurred... It's been removed.

Greelan:
> I am not sure how that occurred... It's been removed.

You haven't committed the change.

Greelan (approved):

PR #804, opened 2024-05-25
Update valheimconfig.json

Goeny:
With the latest update of V+ there are more config options available. This commit adds the Eitr Refinery Settings. Not tested in AMP but as these settings are straight forward i don't expect issues.


IceOfWraith (approved):
Thank you for this! Sorry for the slow response. There were some bugs with a release that prevented it from being tested at the time.

PR #806, opened 2024-05-27
V Rising Add API Port

IceOfWraith:
This adds the port for the API to function on.


Greelan (commented):
Will this actually work for existing instances that don't have the API port allocated - will the manifest cause "$APIPort" literally, or some other invalid value, to be put in the server config file?

IceOfWraith:
I don't expect it to work on existing ones, no. I mentioned that in the development channel.

Greelan:
Won't it break existing instances where someone is trying to configure it manually? Because it will always overwrite what they have included

Greelan (dismissed):
Approved with caveat

IceOfWraith:
I'm going to have to do some thought into this one. Might have to wait until Port refresh logic is hashed out. It doesn't "break" the game server fully, but it does remove the line from the config since it's blank.

PR #877, opened 2024-07-14
Add RAGE:MP

Greelan:


IceOfWraith (approved):
Might make some enterprise users very happy!

Greelan:
I'm trying to get confirmation of the default values for four settings - the wiki doesn't mention those defaults. Unfortunately the RAGE:MP community seems like one of those unhelpful communities where no-one is prepared to give a useful answer to a simple question. Their "solution" is to remove the settings entirely but that of course doesn't work in AMP. I'm assuming setting them to 0 gives the outcome, but haven't yet confirmed.

IceOfWraith (approved):

IceOfWraith (approved):

PR #878, opened 2024-07-14
make v rising config dumber

Allar:


Allar:
Submitted this pr against the wrong depo, doh

PR #892, opened 2024-07-22
Update node.kvp

ssfdre38:
Match name with all other language app runners


Greelan (commented):
You also need to change App.DisplayName

ssfdre38:
Done

IceOfWraith (approved):

PR #916, opened 2024-08-02
Added Config for Seven days to die 1.0 (draft)

Xentraxx:
Config updated, New Quest limit field added and warning for opening Telnet port to Internet added (not necessary).


Xentraxx:
Will do new pull request after more testing.

PR #917, opened 2024-08-02
Updated configs for Seven days to die v1.0 (draft)

Xentraxx:
Updated configs to support v1.0.
For some reason however, the seven-days-to-diesettingsv100.xml is always empty after the initial download of the files.
No idea where that comes from. When I put my config content into the file manually however, everything works like a charm!


IceOfWraith:
That's likely because you have the UpdateSource set to the CubeCoders repo. So it fails. If you set it to yours temporarily it will likely work.

Xentraxx:
> That's likely because you have the UpdateSource set to the CubeCoders repo. So it fails. If you set it to yours temporarily it will likely work.

That was it. Thanks. Now everything works. The Server starts up without any errors but I still can't connect to it....
But I'm assuming thats a networking problem on my side.

Xentraxx:
@IceOfWraith So I am trying around with the port forwarding and what I don't get is that the ports configured in the configv100.xml and the ...ports.json are different than the ones which show up in the AMP Web Interface. It should be port 26900, but in AMP it's 27015. Where is that mapping coming from and where can I configure it? I assume the problem is, that the ports in the Port range 27000+ are being used by Steam and that creates some problems?

Greelan:
Coz those ports are not in your default application port ranges in amp. Change it to "all except excluded"

Xentraxx:
> Coz those ports are not in your default application port ranges in amp. Change it to "all except excluded"

Gotcha! Now the "correct" ports are used. I still get "Could not retrieve Server information" when I try to connect...
I'm testing on the local Network. Any idea how to troubleshoot this further? Local firewall on my Server is off.

Xentraxx:
Okay now it's working! It was the port forwarding in my docker configuration (I use AMP dockerized).
I'll create a new final pull request.

Greelan:
> I use AMP dockerized
Now you know that is not a supported setup

PR #919, opened 2024-08-03
Updated Seven days to die to 1.0 config

Xentraxx:
Tested and working.


Greelan:
Really needed to make all the unnecessary whitespace changes? Lol

Xentraxx:
@Greelan I think this Template is kind of a mess because of all the different versions it tries to handle.
I believe it's quite confusing. Like which setting to use for which version.
What do you think about creating a completely new template "Seven days to die" and renaming the other one to something along the lines "Seven days to die pre-release", to have a new start and a fresh and tidy template?

IceOfWraith:
Rather than that I think it'd be better to just put the settings into a different subcategory.

PhonicUK:
I've got a pending PR for the new subcatetories, but I don't want this PR accepted with the formatting changes showing it as "Whole files removed/while files added" because it makes it impossible to review the changes.

IceOfWraith:
If you can fix the spacing to 4 spaces in the JSON files I'll review this one. As is, it's too messy and doesn't match the formatting of others. There's a setting in VS that adjusts this.

Xentraxx:
I tried, but even though I changed the settings in VS code and the files look fine for me, VS code does not recognize this change as a change of the file, so apparently I can't save it.

Greelan:
@IceOfWraith I have included some spacing fixes that should make it easier to review

IceOfWraith:
Thank you! I'll look over it.

Greelan:
Tho now there are conflicts due to the previous reconfiguration of the settings categories/subcategories.

Might be easier for the OP to start over tbh

PR #930, opened 2024-08-07
Add Spellmasons

Chain-BC:
Spellmasons dedicated server module partly based on the existing nodejs module, as it requires node to host the server.


Chain-BC:
A bit of context as to why I did some things the way I did. Currently Spellmasons servers do not store any data (any game is saved on the client side) hence why the whole game folder is deleted when updated. I also am just a bit inexperienced with this kind of stuff, so that is part of the reason as well. Also, the only way to get the server is to build it from the GitHub repository or possibly get it through docker, but I haven't used docker very much so I opted for GitHub instead. If the module needs work and needs a rewrite do let me know, I will close the pull request or make it a draft, any feedback would be appreciated as well. If someone else wants to write a better one, that is fine too.

IceOfWraith:
I'm not familiar with this game, so haven't looked too closely yet. I'll try to get to it this week.

Greelan:
Has this been tested on Windows? `cp` is not a Windows command

Chain-BC:
I tested it on Windows and Linux, and both worked just fine but perhaps it may be missing some file needed for modding, I am double checking and rewriting what I need to rewrite so I'll make sure it works.

Chain-BC:
Any update on this? I should have gotten it to work on Windows with no problems now but I may have missed something.

IceOfWraith:
What's the reasoning behind pulling a specific version by default versus getting the latest?

Chain-BC:
I haven't figured out how I should do this yet, but currently the latest version tends to be the beta version which you have to opt-in for on steam. So my reasoning is pull the specific latest stable by default instead of latest beta, since most people won't be using the beta version. A bit of a hassle but I can't think of another way to do it.

IceOfWraith:
Okay, that makes sense. Kind of a pain but nothing you can control.

Chain-BC (commented):
Everything seems much more well organized and the changes just made it overall better, and after testing it out everything works, so I am not opposed to the changes.

Chain-BC:
The developer has been messing around with improving servers, but I don't believe it changes how dedicated servers are hosted so it should be fine.

Greelan (approved):

PR #946, opened 2024-08-15
Terraria & Tmodloader: update command line args

Greelan:
Works around the current AMP bug


Greelan:
Closing because Ice is too slow and #949 did the same

PR #950, opened 2024-08-15
Add TES3MP

hmtack:
First time creating a configuration. I referenced the Skyrim Together Reborn configs as the two multiplayer mods share a lot of server configuration settings.

[TES3MP](https://github.com/TES3MP/TES3MP) is a multiplayer fork of [OpenMW](https://github.com/OpenMW/openmw), which is itself an open-source version of The Elder Scrolls III: Morrowind. 

I've confirmed the current configuration creates a working server on my own Debian server with Wine. There are still a few issues I'm checking out:

- Not 100% sure the server list port (25561) has to be ported for the server to appear on the browser. I'll test a server without it forwarded.
- The active users widget doesn't reflect the maximumPlayers setting. Will need to figure out how to link them.
- The console returns "Could not complete action" once per server start. This has seemingly no affect on the server functionality.


Greelan:
> * The active users widget doesn't reflect the maximumPlayers setting. Will need to figure out how to link them.

Use $MaxUsers in the FieldName for the setting

> * The console returns "Could not complete action" once per server start. This has seemingly no affect on the server functionality.

You need to configure the regex for ChildProcessName. Look at `ps -aux` to figure out the right regex. Look at any other Wine template for how it might be written.

PR #963, opened 2024-08-17
Improve file extensions

Bluscream:
Template file extensions were always kind of weird, and i know they're arbitrary, but it just makes more sense to follow some pattern, thats why the whole `*config.json` and `*metaconfig.json` stuff happened in the first place. However, the current way without any delimiter, makes it harder to programmatically go over the different config types (`*metaconfig.json` also contains `*config.json`, y know) and it's generally just harder to read when you look over a repo and see this
librewolf_DtkdTIP1Fe

I would suggest; for the future atleast, you enforce such a naming scheme for Pull Requests that want to be merged into the official template repo


IceOfWraith:
I'm closing this since it would be a breaking change as we found out. A fix has been put in to address it, but old versions of AMP before the fix would error with these changes.

The plan is to combine templates into one JSON file in the near future which would overcome this problem as well.

PR #979, opened 2024-08-25
NightOfTheDead: Allow the setting UnlockAllBuildings to be set in config file

sunerievers:
According to server documentation, and local testing, it is possible to configure so that all buildings can be created even if not unlocked yet in main story. This allows for a more QOL experience, and I would like it to be available as a setting for the AMP template for the game.

https://steamcommunity.com/sharedfiles/filedetails/?id=2252239447

### DIFF
--- a/night-of-the-deadconfig.json
+++ b/night-of-the-deadconfig.json
@@ -965,5 +965,21 @@
             "240":"240",
             "300":"300"
         }
+    },
+    {
+        "DisplayName":"Unlock All Buildings",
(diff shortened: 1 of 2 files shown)


Greelan:
Thanks. This will need to be done a slightly different way to avoid breaking existing instances that have not yet been updated. I will get around to making those changes in the next few days

Greelan:
A lot has changed since the template was created, hence #982

PR #1017, opened 2024-09-20
Added Custom Server Startup Command Field to Abiotic Factor

leakedleague:
Hello there, 

I added a custom Server Startup Command Field to Abiotic Factor. This should give some flexibility for custom options and fixes in the future.

Best,  

leakedlague


Greelan:
No issues in principle. Only query is the placement. Normally I aim to put the custom field in a place that means that if someone puts something in like Port that is already configured by AMP, AMP will still prevail. With this server, does first win or last win?

leakedleague:
Its first win, thats why i placed the field afterwards, you cannot overwrite the Port set by AMP with the field, you still have the Option to add other commands though, which would give more flexibility.

Greelan:
Right, needs to be after ServerName etc then

leakedleague:
the bigger Problem i see is you put this: {{useperfthreads}}{{NoAsyncLoadingThread}}. Technicly not bad with bad hardware, useallaviablecores is with this server better though with good hardware

Greelan:
That would have been based on the server instructions at the time

Greelan:
https://github.com/DFJacob/AbioticFactorDedicatedServer/wiki/Technical-%E2%80%90-Launch-Parameters

leakedleague:
As im kinda new to AMP, is there any possibility to implement else or? So for example you cannot check the option "NoAsyncLoadingThread" when "Useallaviablecores" is initialized?

There is literally no Documentation about stuff like this

Greelan:
There is not that possibility. Just have to explain it in the settings

leakedleague:
Alright before i close this Pull request to add more things, i know that you can set async and streaming parameters bound to the Engine itself in another config file thats not beeing created nativly. Its the Configuration file Engine.ini, which currently  needs to be created manually in the same directory that Sandboxsettings.ini gets initialized.

My question would be does AMP generate those Inis (in abiotic Factorconfig and Metadata Sanboxsetings.ini is described) or does it just access existing conigurations and changes them. Is it possible for me to create an ini file that before did not exist or not?

leakedleague:
Just to add what exactly i would add there:

[/Script/Engine.StreamingSettings]
s.AsyncLoadingTimeLimit=40
s.ForceGCAfterLevelStreamedOut=0
s.LevelStreamingActorsUpdateTimeLimit=10
s.PriorityLevelStreamingActorsUpdateExtraTime=10

Greelan:
You don't need to close the PR to add commits.

AMP doesn't automatically create configs or write stuff into configs that isn't there already. For those Engine.ini settings, you could cheat by using -ini flags on the command line

leakedleague:
Can you tell me a quick example how you mean int? Because i am not sure how you mean command line flags. 

These settings are not applied and read by abiotic factor but by the Engine itself. you also do not need to load Engine.ini, its enough if you create it. As soon as it exists, the settings inside are the highest priority settings overall. 

For example if you add :
[URL]
Port=25994

it will automaticly overwrite any startup settings and ignore the port you gave it in startup. 

so i am a bit conufsed. Sure i could probably chain the creation of the Ini file into the startup as a script but that would be tbh beside the point.

Greelan:
An example: https://github.com/CubeCoders/AMPTemplates/blob/b6aa40a8d3f07367230fb434049959a34f9cf272/satisfactoryconfig.json#L41

This is then added to the command line via {{$FormattedArgs}}

You really don't need to keep closing PRs just to make further changes...

PR #1018, opened 2024-09-20
Abiotic Factor: Addition of optional Servercommands, addition of option useallavailablecores in Server Config

leakedleague:
Hey there, 

for now my final contribution to config. Added the option to insert custom Server Commands. Field is added at the end of startupline and should not compromise AMP Port config at all, even if someone tried to mess with it. 

Additionally the Performanceoption "useallavailablecores" in Serversettings got added. It is good to reduce the Loading time on newer Serverhardware. 

This commit works flawlessy on my AMP Setup, if any Improvements can be done feel free to send a message. 

Best, 

Leakedleague


Greelan (approved):
Approved with fixes

PR #1030, opened 2024-09-28
OS_CLOSE causes eco process to close without fully saving

NHDriver4:
Replacing App.ExitMethod=OS_CLOSE with App.ExitMethod=CtrlC allows the Eco game process to properly shut down. OS_CLOSE does not allow full game saves as the process seems to terminate early.


Greelan:
Doesn't shut down properly on Linux. So you will have to wait for the generic module to allow different shutdown methods per platform

PR #1047, opened 2024-10-01
Update minecraft-bedrockdownloadurl.json

ssfdre38:
Bedrock Download URL Change. New download links added to download template


ssfdre38:
Changing update stage to match new urls. New pr will come

PR #1063, opened 2024-10-10
Proton: revert update stages

Greelan:


IceOfWraith:
I'm still getting the end of line escaping error using this version. Can you confirm it's actually working on your end?

Greelan:
Jesus - this actually works:

`grep browser_download_url | cut -d\\\\\\\" -f4 | grep .tar.gz`

Greelan:
And it addresses the third level of nesting in other cases where a single quote needs to be changed

IceOfWraith (approved):

PR #1080, opened 2024-10-19
Seven Days: add 1.0 settings

Greelan:


IceOfWraith (dismissed):

Greelan:
If you want to remove the backup exclude, you also need to remove the update stage

IceOfWraith:
Done. Forgot we had that as an option also.

IceOfWraith (approved):

PR #1086, opened 2024-10-22
Various: update tar commands

Greelan:
Rely on $PATH rather than assuming location

### DIFF
--- a/factorioupdates.json
+++ b/factorioupdates.json
@@ -50,7 +50,7 @@
         "UpdateSourcePlatform":"Linux",
         "UpdateSource":"Executable",
         "UpdateSourceData":"/bin/bash",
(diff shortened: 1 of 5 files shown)


IceOfWraith (approved):

Maxfojtik:
This commit prevented the instance from extracting when doing a fresh install

Greelan:
As the `amp` user, what is the output of:
`which tar`
`echo $PATH`

Also, is the instance in docker?

Maxfojtik:
/usr/bin/tar
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
yes it is in a docker

Welp, I went back to the "CubeCoders/AMPTemplates" Configuration Repository to try to replicate the issue and of course it started working. It was complaining about not having an end quote on the command for extraction. 

https://discourse.cubecoders.com/t/update-failure-with-factorio-factorio-update-failing/18196

this is a forum post describing the issue I had created by someone else

Greelan:
Yes, that was related to the .NET 8 update and its ability to handle escaping 

That's why the release notes said you needed to update your instances from ADS to refresh the template config

PR #1101, opened 2024-10-31
Update ark-saconfig.json

Gomezb3:
added dino event colors


Greelan:
See comments on #1102

PR #1102, opened 2024-10-31
Update ark-sa.kvp

Gomezb3:
Added support for colors


Greelan:
All related commits should be part of the same PR

Also, the ASA minimal template needs the same changes

PR #1159, opened 2024-12-02
Add support for Fistful of Frags

connorrichlen:
as titled


Greelan:
Could you change the root dir to fof rather than srcds for consistency with other templates? Have all the setting been tested - everything srcds server is different

Greelan:
Thanks for the changes. Why is the template labelled Experimental?

Greelan (approved):

IceOfWraith (approved):

PR #1191, opened 2024-12-20
Various: Update Metamod download option description URL and CS2 default version

colehammond65:
Changed the URL in the Metamod branch and build version settings as Metamod now uses metamodsource.net instead of sourcemm.net for their downloads page.
Also updated the default build version to 1315 from 1313 to keep it up to date


Greelan (approved):
Thanks. May as well do them all tho xD

PR #1200, opened 2024-12-23
Add Left 4 Dead

IceOfWraith:


IceOfWraith:
This game doesn't appreciate higher than normal ports. 41000 was too high and caused it to revert to a 7-8000 range port.

PR #1203, opened 2024-12-23
Update factorio.kvp so Console.UserChatRegex= works correctly.

mindcloud69:
Previous Console.UserChatRegex was not javascript regex compliant.  Fixed and tested working on 2.0.28 of Factorio.


mindcloud69:
I also.

* added the colon after the username so that it does not get capturn in the username field. 
* made the gps capture non greedy to capture the closing ] correctly.
* Changed the last capture to * instead of + because there are situations in Factorio where a blank message chat can be potentially sent. Edge case but why not account for it.

Greelan (approved):

PR #1208, opened 2024-12-26
Critical: Update dotnetcli domain

richlander:
`dotnetcli.azureedge.net` may fail soon.

FYI: `dotnetcli.blob.core.windows.net` is the storage account this CDN. Best to use the new CDN.

Related: https://github.com/dotnet/core/issues/9671


IceOfWraith (approved):
Thank you for the early notification! Happy holidays to you and the .NET Team!

PR #1228, opened 2025-01-06
Add VC2017 to space engineers wine script (to fix server not starting in Docker)

emidah:
Hi!

This PR fixes Space Engineers instances not starting when using Linux (Docker).

### How and why:

I use AMP with Debian 12 + Docker.

For me, every clean instance of the Space Engineers server would crash with some cryptic DLL error no matter how many times I recreated the instance:

`Exception while loading world: An error occurred while loading the world.`

In the log:

`ERROR Entity init!: System.DllNotFoundException: Unable to load DLL 'VRage.Native.dll': Module not found. `

So I investigated online, and found that the server needs VC2017.

After I added it to the script, the instances immediately started working.

### DIFF
--- a/space-engineers-genericwinescript.sh
+++ b/space-engineers-genericwinescript.sh
@@ -23,6 +23,7 @@ export DISPLAY=:$DPY_NUM
 ./winetricks corefonts > winescript_log.txt 2>&1
 ./winetricks sound=disabled >> winescript_log.txt 2>&1
 ./winetricks -q vcrun2013 >> winescript_log.txt 2>&1
+./winetricks -q vcrun2017 >> winescript_log.txt 2>&1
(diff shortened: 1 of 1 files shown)


IceOfWraith:
Thank you for the pull request. I tested on my end and it still gave the error. Let me do some more testing and we'll figure out what's going on.

emidah:
FYI the update.json (or something like that?) is hardcoded to pull the newest script from the main branch of this repo. So just changing this file and setting it as the amp template source is not enough, but you need to also modify that file

IceOfWraith:
Haha yeah... I overlooked that and Greelan also called me out on it. I'm just waiting for the test instance right now. This is a slow one to test.

IceOfWraith (approved):
All good here. Thank you for addressing that!

PR #1273, opened 2025-01-26
Add Puck

CodionCA:
The dedicated server on Steam is Linux only, is there something more needed for Windows users? Works fine dockerised.

https://store.steampowered.com/app/2994020/Puck/


Greelan:
Thanks for the contribution. I've made a bunch of changes to make improvements and tidy stuff up (the config generator has a few quirks and issues)

Let me know what you think

BTW, normally the steam connect link uses the Query Port, rather than Game Port. How does this server operate?

Greelan:
Is there a direct connect option in-game or is it only via the server list/browser?

CodionCA:
Its only in the server browser I believe, unless you used some sort of extra debugging tool. 

From the devs discord
From clients perspective, here is what happens:
1. Game is loaded, WebSocket is connected, steam authentication is performed
2. Client presses refresh button
3. Backend provides the client with list of servers along with their IPs and `--port`s.
4. Client iterates over each server in the servers list and opens UDP socket with server's `--port`+1. So `7778` for our scenario.
5. Client sends `ping` message to this port and waits for response with a timeout of `250ms`
6. If `pong` response is received, the time between send and receive is measured and used as displayed ping and server is displayed in the server browser UI
7. If no response (within `250ms`) is received, server is skipped and not added to the UI server browser at all.

Greelan (approved):

PR #1293, opened 2025-02-06
Update Necesse Configs

NiklasRosenkranz:
Added the option to set the Server Owner in the Configuration


NiklasRosenkranz:
hey, 
yeah a little oversight on my part it isnt required but has to be in commandline otherwise it didnt work on my test server

Greelan:
What's the point of including it in the template file then?

Are you sure that your test instance was downloading your modified template file?

NiklasRosenkranz:
the reason i put it in the file is that there should be a "easy" way to set the Server Owner without editing the start file. 

and yes the modified file was downloaded and loaded on startup

Greelan:
I disabled the metaconfig to see what the server itself creates. Doesn't have any owner entry:

```
ERVER = {
	port = 14159, // [0 - 65535] Server default port
	slots = 10, // [1 - 250] Server default slots
	password = , // Leave blank for no password
	maxClientLatencySeconds = 30,
	pauseWhenEmpty = true,
	giveClientsPower = true, // If true, clients will have much more power over what hits them, their position etc
	logging = true, // If true, will create log files for each server start
	language = en,
	unloadLevelsCooldown = 30, // The number of seconds a level will stay loaded after the last player has left it
	worldBorderSize = -1, // The max distance from spawn players can travel. -1 for no border
	droppedItemsLifeMinutes = 0, // Minutes that dropped items will stay in the world. 0 or less for indefinite
	unloadSettlements = false, // If the server should unload player settlements or keep them loaded
	maxSettlementsPerPlayer = -1, // The maximum amount of settlements per player. -1 or less means infinite
	maxSettlersPerSettlement = -1, // The maximum amount of settlers per settlement. -1 or less means infinite
	jobSearchRange = 100, // The tile search range of settler jobs
	zipSaves = true, // If true, will create new saves uncompressed
	MOTD =  // Message of the day
}
```

Looks like it is actually only a command line argument:

https://necessewiki.com/Multiplayer#Server_Parameters

Looks like the whole template could do with a lot of updates

Greelan:
@NiklasRosenkranz not sure why you closed the PR, but I've prepared my own updates to the template to add all the settings (command line and server.cfg). The existing template (not mine) has -seed and -guide as two command line parameters, but they are not listed on the wiki, and I can't find anything about them

Do you know if they actually work?

PR #1297, opened 2025-02-09
Passive Mods Support

Nypheena:
[The Great Ark Mod Settings Debate (or not)](https://discord.com/channels/266012086423912458/1334993662538682529)


Greelan:
Thanks for this. Couple of observations:

- The change to the Use Mods setting won't kick in for an existing instance that already has the setting on true, until the user toggles the setting. Unavoidable, but that might cause some confusion
- Would it be worth adding a comment to that setting that the Passive Mods list doesn't need to be repeated under Mods List? And/or change the Mods List setting description to "Active Mods" or somesuch?

Greelan:
Will the server crap itself tho if -passivemods= has a mod list, but that list doesn't appear in -mods=?

Greelan:
Right, so effectively `-passivemods=` will override whatever is in `-mods=` so even if the mod is specified in `-mods=` then  its functionality will still be disabled?

And it doesn't matter whether the mod is in `-mods=`, since if it isn't, `-passivemods=` will do the work to add it to the mod list?

Greelan:
Gotcha, I was just thinking about adding some extra commentary to the passive mods setting to describe the behaviour (ie it will have its effect regardless of whether the mods is in `-mods=`)

Greelan:
If you are happy with my commit just now, I will approve and merge this

Greelan:
Thanks for doing the testing on this

Greelan (approved):

PR #1309, opened 2025-02-14
ARK: SA Add Astraeos as a map

Gomezb3:
Added Astraeos as a map to both relevant configs. The map isn't released yet but is supposed to be in a couple hours.


IceOfWraith:
Thanks! Can you please bump the version numbers in the kvp files as well?

Greelan (approved):

PR #1316, opened 2025-02-17
ARK:SA Unified Save Data (usestore)

Nypheena:
This adds support for UseStore, ConvertToStore, and BackupTransferPlayerDatas, enhancing clustering and server transitions while preventing data loss. It also improves integration with ARK’s built-in backups, making it easier to roll back with character and tribe data in sync, as all information is stored within the map file that the game automatically copies.

Per Patch Notes:
Additional Server related params/config options for using a “usestore” save system which combines player profiles, tribe data, and the world save into one single file. 

Note: This is not new, but not previously documented to our knowledge and what we use on our Official Network.

-converttostore (imports the current player and tribe data into the .ark save file)

Run with -converttostore -usestore and then the next time you relaunch, you should not need -converttostore as all the data will be integrated.

Important to note: .arkprofile & .arktribe files are not automatically deleted after conversion so should be deleted manually if you want to save storage space.

-usestore (utilizes a different saving method that stores player and tribe data inside of save files)

Important to note: when rolling back save data simply do as normal and all the arkprofile and arktribe data will be in sync with the backup save already.


Greelan:
I reckon it gets lost in the Clusters setting, and people will assume it is only relevant to Clusters

Greelan (approved):

PR #1342, opened 2025-03-03
Sven Co-op: various fixes and updates

Greelan:


IceOfWraith (approved):
This brought back so many memories. Good times...

PR #1369, opened 2025-03-20
Beasts of Bermuda: switch to Wine

sanegaming:
Revert back to using Windows Binary via Wine, Linux binary is not updated by the developers properly.


Greelan:
You will also need to changes the updates.json file

Greelan:
It should also use the "wine8" image (which actually has Wine 9 in practice), and I suspect the child process regex is not right

Greelan (approved):

PR #1392, opened 2025-04-02
Update arma-reforgerserver1-5.json

jhendrix99:
Changes the public address to be the same as bindAddress. This fixes it binding to wrong IP when using multiple IPs on a machine.


Greelan:
Not the right solution. The right solution is waiting in #1156

PR #1396, opened 2025-04-05
Killing Floor 2: Added support for custom command-line arguments

arukaraz:
## Description

This PR adds to the KF2 configuration additional startup arguments: `CustomUrlArgs` and `CustomFlags`.
## Changes

- Added `CustomUrlArgs` for URL-style parameters (e.g., `?Mutator=KFMutator.MyMutator`).
- Added `CustomFlags` for dash-prefixed flags (e.g., `-NoSteamClient`).
- Ensures AMP-defined flags always take precedence over user-defined ones.
- Improves flexibility and compatibility with mods, mutators, or other custom server configurations.

### DIFF
--- a/killing-floor-2.kvp
+++ b/killing-floor-2.kvp
@@ -16,7 +16,7 @@ Meta.ContainerPolicyReason=AMP's console is unable to provide console output on 
 Meta.Prerequisites=[]
 Meta.ConfigReleaseState=NotSpecified
 Meta.AppConfigId=75f76d07-e6a2-4b45-868a-8fc7e9bea137
-Meta.ConfigVersion=1.1
+Meta.ConfigVersion=1.2
 App.DisplayName=Killing Floor 2
 App.RootDir=./killing-floor-2/
 App.BaseDirectory=./killing-floor-2/232130/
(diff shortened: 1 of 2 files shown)


Greelan:
Question: if someone includes in CustomArgs a URL or dash style argument that AMP already sets, what happens? Error? First prevails? Second prevails? This is important to deciding where to place the variable in the command line

Greelan:
The other issue I have is that by combining both styles in the one argument, it presumably won't work if only a dash style argument is set, given there is no leading space

arukaraz:
Hey @Greelan , sorry for the late reply.

Regarding your first question:
If a user includes any command that's already defined by AMP (e.g., -Port or -QueryPort), the game will still run without errors. However, this will conflict with the ports defined by AMP, and it seems they effectively cancel each other out. So it's important that the user knows exactly what they’re doing in such cases.

As for your concern about leading spaces:
It turns out that if you include a leading space, it will be reflected in the command line. So if a user wants to add just a single dash-style argument, they need to start with a space before the argument.

But let me know if you'd like me to add more clarification in the description

Greelan:
I don't understand the first response - if the server doesn't error, usually one or other will prevail. That should then determine where the {{CustomArgs}} is placed in the command line, to ensure that AMP's settings always win

Requiring the user to add a space is too finicky. The arguments should be separated - one variable for URL args, one for dash args. There are various other templates that take this approach, such as Ark

arukaraz:
You're absolutely right. I'll go ahead and push the changes using that approach—separating the arguments into two variables. That definitely makes things cleaner and more user-friendly.
Thanks for the feedback!

arukaraz:
@Greelan I just split the arguments into separate fields. AMP-defined flags will now always take precedence over user-defined ones.

Greelan:
I've made some changes for consistency with other templates, and also to reflect the ordering that my testing suggested is needed to ensure the custom options don't override AMP's settings

Greelan:
@arukaraz any further comments?

Greelan (approved):

PR #1415, opened 2025-04-20
[Old] Added SRB2

Hacksawfred3232:
My attempt at trying to add SRB2 to CubeCoders AMP. It... works, albeit clunky. See https://github.com/Hacksawfred3232/CustAMPTemplates/issues/1

I've tried to add support for the Windows binary, but it seems insistent on opening a separate window for the console, which means it will most likely not work with AMP. So only the Linux version is supported, despite the only distributed Linux binary is Flatpak, which means it has to be compiled.


Hacksawfred3232:
Well, probs still clunky, but at least it works. First attempt at this, edits welcome.

Hacksawfred3232:
Due to wanting to work on other templates that can be merged into the main branch, I will go ahead and close this pull request so I can reorganize the forked branch.

PR #1416, opened 2025-04-21
Add H1-Mod (Call of Duty 4 Modern Warfare Remastered)

Draakoor:
Template only works on windows, because binarys are not supported on linux (no proton, no wine)


Draakoor:
Template works now
image

PR #1420, opened 2025-04-23
Add H1-Mod and HorizonMW

Draakoor:


IceOfWraith:
Templates made with the config generator won't be accepted without major refactoring. This one doesn't follow the consistency of other templates in the repo and has no Linux version added. If it's cleaned up to match other templates it can be reviewed again.

PR #1429, opened 2025-04-28
Terratech worlds config update

Airfirewall:
Server settings and config updates

- Added a config template due to one being downloaded from steam is no longer up to date with the new server settings.
- Removed tech yard storage multiplier because it is no longer a valid server setting.
- Updated Tech Reactor Capacity to include the Infinite reactor size.
- Added 3 new game settings;
		1. Enable Structure Power Requirement, Can enable or disable power requirement for structures.
		2. Block Licensing - Can enable or disable requirement for parts to be licensed before it can be stored in tech yard.
		3. Blocks Globally Accessible Without Tech Yard - Can enable or disable the ability for players to access tech yard storage without having to be at a tech yard.


Greelan (approved):
Approved with tidy-ups

PR #1435, opened 2025-05-01
Sons of the Forest: Add logging exclusion

KairuByte:
Ignore common outputs that read like errors but mean nothing to the end user


Greelan:
One small nit - presumably the dot in `Physics.ClosestPoint` should be escaped?

KairuByte:
> One small nit - presumably the dot in `Physics.ClosestPoint` should be escaped?

I unironically love nitpicks, good catch. Corrected ^_^

Greelan (approved):

PR #1454, opened 2025-05-11
Add SRB2

Hacksawfred3232:
Now using branches instead of sticking everything into main. This should allow me to go and deal with some requests that I should be able to deal with.

These templates are also available in [Hacksawfred3232/CustAMPTemplates](https://github.com/Hacksawfred3232/CustAMPTemplates), so if anyone wants to test them out before they get merged into main, they can do so.

I've tried to add support for the Windows binary, but it seems insistent on opening a separate window for the console, which means it will most likely not work with AMP. So only the Linux version is supported, despite the only distributed Linux binary is Flatpak, which means it has to be compiled.


IceOfWraith:
Haven't had much time to review how this one was all setup. Just a heads up that there may need to be some changes to how it's done if it's officially added to AMP. I'll look at it soon and see what you've got going and get a review to you.

PR #1479, opened 2025-05-21
Added ability to import some settings

Gomezb3:
Replicated 20 settings to allow for importing. I took a look at Ark SA's config and went from there and I believe everything is working as it should. As a test on my own instance I replicated the day/night cycle speed multiplier and metaconfig and I was able to successfully import changes.


Greelan:
Couple of comments:

- This should be extended to other settings that are able to be set by the user in AMP's UI but are currently set by command line option only, yet also have a ini equivalent. ASA is different to ASE, since with ASA most ini settings stick, whereas with ASE a lot of the settings are stripped from the ini when they are on the default value. The exceptions to that for ASA were the ones with the backup command line option (which you have basically copied).

  If all relevant settings are not included as importable, then this will just prompt support queries.

- The hidden setting for each should be the one that writes to the config - so the reverse of ASA. The ASE approach is basically to use command line options for almost all settings (if possible), given the behaviour noted above, and so the hidden setting is just to allow importing. Hidden settings should come second. ASA has the opposite approach - all settings are set by writing to the inis, with only a few having a backup command line option.

  Example (basically switching the ParamFieldName and IncludeInCommandLine settings):
  ```
    {
        "DisplayName": "Day/Night Cycle Speed Multiplier",
        "Category": "ARK SE:stadia_controller",
        "Subcategory": "Multipliers:tune:3",
        "Description": "Sets the scaling factor for the server day/night cycle. The default value 1 sets the same speed as a singleplayer game, ie 1 minute real time = approx 28 minutes server time. A scale of 0.035 would set a day/night cycle of approx 24 hours real time",
        "Keywords": "day,night,cycle,speed,multiplier,daycyclespeedscale",
        "FieldName": "DayCycleSpeedScale",
        "InputType": "number",
        "IsFlagArgument": false,
        "ParamFieldName": "DayCycleSpeedScale",
        "IncludeInCommandLine": true,
        "DefaultValue": "1.000000",
        "Placeholder": "1.000000",
        "EnumValues": {}
    },
    {
        "DisplayName": "Day/Night Cycle Speed Multiplier",
        "Category": "ARK SE:stadia_controller",
        "Subcategory": "Multipliers:tune:3",
        "Description": "",
        "Keywords": "",
        "FieldName": "DayCycleSpeedScale",
        "InputType": "number",
        "IsFlagArgument": false,
        "Hidden": true,
        "ParamFieldName": "ServerSettings.DayCycleSpeedScale",
        "IncludeInCommandLine": false,
        "DefaultValue": "1.000000",
        "Placeholder": "1.000000",
        "EnumValues": {}
    },
    ```

Gomezb3:
Thank you, I am willing to do the rest for the other settings but I wanted to make sure I got it right first before doing everything. To make sure I fully understand, for the settings I have replicated I just need to swap the ParamFieldName and IncludeInCommandLine so that the hidden config includes ServerSettings (while the default one does not) and to make sure the hidden one is not included in the command line? Also, for the hidden config does DefaultValue and Placeholder need to be included? In the ASA config I saw that the hidden configs did not have it so I left it out as well, but I wanted to double check on that. Also, for any settings that pertain to game.ini, (specifically Structure Damage Repair Cooldown) what would I add to the ParamFieldName? I'd assume that it isn't ServerSettings since that is a GUS.ini entry, would it be /script/shootergame.shootergamemode? Thank you!

Greelan:
> To make sure I fully understand, for the settings I have replicated I just need to swap the ParamFieldName and IncludeInCommandLine so that the hidden config includes ServerSettings (while the default one does not) and to make sure the hidden one is not included in the command line? 

Yes

> Also, for the hidden config does DefaultValue and Placeholder need to be included? 

No

> Also, for any settings that pertain to game.ini, (specifically Structure Damage Repair Cooldown) what would I add to the ParamFieldName? I'd assume that it isn't ServerSettings since that is a GUS.ini entry, would it be /script/shootergame.shootergamemode? Thank you!

Yes, include the section heading (case sensitive). There are a couple of examples already in the config manifest

Gomezb3:
Thanks for the quick response, currently I have all applicable settings in the Multipliers, Structures, and Clusters tab able to be imported which is pretty much everything there except structure stasis grid, cluster directory override and cluster ID since I believe those are all only command line options. Tested all of the Structures settings and a few multipliers settings and had no issues getting them imported. I will have time to do the rest of the applicable settings in the Gameplay and Server tab tomorrow. Thank you!

Gomezb3:
I believe I have finished adding importability for the rest of the settings that aren't purely command line options. I did also convert flyer speed leveling from a command line option into one that uses the ini, but I noticed that it wouldn't change the state from true to false (or vice versa) unless the ini line was already present. Once I went ahead and added the line to game.ini, I was able to freely change it and import it without issues.

Greelan:
@Gomezb3 why did you close this? Don't tell me I wasted my time reviewing and editing it....

Greelan:
You should be able to reconstruct it from the commit history. Go to the file state at my last commit

Gomezb3:
Don't see any file states below your last commit or commit history in this PR, but I was able to find [this](https://github.com/CubeCoders/AMPTemplates/blob/5e5c4e9cd98a23caf4856df074d695d08e1a94ed/ark-seconfig.json)
I believe that is the one you are referring to?

Gomezb3:
After reading it a bit, I think that is the correct one, I found it from commit [5e5c4e9](https://github.com/CubeCoders/AMPTemplates/commit/5e5c4e9cd98a23caf4856df074d695d08e1a94ed) which has the commit [200510b](https://github.com/CubeCoders/AMPTemplates/commit/200510b0467b12ab8856934e43b53c1f41ee6bcc) as its parent. That second commit was the one where you tidied up a lot of the changes and provided some pointers. Gave it a quick look through and it does add a lot of hidden options to allow for importing practically all UI settings.

Issue #1301, opened 2025-02-10
[Game Request] Aloft


capi1992:
# Module Request

Note: Please try setting up a configuration yourself before raising an issue to request a configuration:  ~~https://config.getamp.sh/~~
***There is a newer beta version available at https://iceofwraith.github.io/GenericConfigGen/

## Game Information:

Game Title: Aloft
How to obtain: https://aloft.wiki.gg/wiki/Dedicated_Server_how-to
Suppoted OSs: Windows
Supports RCON/Console input? Powershell

## I confirm:

- [x] that I have searched for an existing module request for this application.
- [x] that the server is an official server distributed by the games publishers/developers.
- [x] that the server is available to the general public.
- [x] that the server can be run headless (without a GUI).
- [ ] I have already attempted to create a configuration myself using the configuration tool
- [x] that my checkboxes above look like the one on the left (remove this line when done)

Greelan:
Already have a template in `Greelan/AMPTemplates:dev`

However, it only runs on Linux, because on Windows the server refuses to generate a world or start when run as a service (which is the default method that game servers are run by AMP on Windows)

PR #1648, opened 2025-08-17
Added Aloft to templates

bradwbu:


Greelan:
See https://github.com/CubeCoders/AMPTemplates/issues/1301

PR #1700, opened 2025-10-12
Subnautica 2.0 Update

Nomlytx:
Nitrox has released version 1.8, which is compatible with Subnautica 2.0.

Could you assist with this update? @Greelan


Greelan:
1.8.0.1 of Nitrox has dropped, but there is a bug in the save path changes, so still waiting for the next release (I raised the bug with the Nitrox developers and they have patched it)

PR #1703, opened 2025-10-13
Update v-risingconfig.json to fix LanMode flag

Choque:
The LanMode flag should correctly be added to the ServerHostSettings.json with a value of true or false.


Greelan:
An alternative command line flag has been implemented

PR #1764, opened 2025-12-11
ANEURISM IV: update maps and settings

CaptRedB3ard:


Greelan:
I've made a few corrections and consistency changes

I notice that the server adds to aniv.cfg:

```
maintenanceTime "null"
remoteRanksURL ""
```

But you haven't added those to the config manifest and I can't find any information on the wiki or the Vellocet discord about them. Any ideas?

CaptRedB3ard:
Ah yes there are a few undocuments options within the server, those two are optional but not needed by most users as maintenanceTime is just the server restart which can be done through the amp scheduler instead, and the remote ranks file is just to specify a URL that the server can load a ranks file for for user permissions configuration which is really only helpful when running multiple servers, it's the system we use on official servers to load permissions. Also to add I am affiliated with Vellocet, this can be confirmed in the discord if you ever have questions you can consider me a point of contact within vellocet

CaptRedB3ard (commented):
functions as intended. thank you for your fast responses and help, much appreciated.

PR #1765, opened 2025-12-12
Eco: Add server remote address configuration

BroOtti:
Added configuration for server remote address with details.

As requested on discord: https://ptb.discord.com/channels/266012086423912458/1448726068763557951


Greelan:
Is that setting in the template config file that is supplied by the server? Presumably it is optional, in that in normal situations where there isn't multiple IPs or a proxy or tunnel, the server finds and reports the external address itself?

if so, I think it is preferable for the setting to have options for server determines, amp determines, and custom, like a number of other templates that have similar settings

BroOtti:
I checked before creating the PR that the setting is in the config file and to ensure I name it correctly :D
Also, I checked on the wiki and borrowed the description https://wiki.play.eco/en/Server_Configuration/Network.eco

I don't know if this setting is optional. I will ask the discord user who requested this setting to be added.

Wouldn't "server determines" and "amp determines" probably be the same? Not sure, how "amp determines" would look like. Iirc, there was something like ExternalIP as variable?
Sadly, I'm not that deep into templates yet. How would you do this or do you have a template in mind which already does this so I can look how to do it?

Greelan:
I only asked whether the setting is there because I know from previously looking at it that not all were

Voyagers of Nera is a recent example. Obv here you would also need to add the port

The first two options won't necessarily be the same in a multi-ip scenario

PR #1784, opened 2025-12-27
Path of Titans: add new gameplay settings for dinosaurs and respawn

sanegaming:
Added:
New Riparia Map to dropdown as it is an official map.
Information for disallowing certain dinos rather than full whitelist system. bServerWellRestedBuff
ServerRespawnTime
bPermaDeath
bDeathInfo


Greelan (approved):
Approved with fixes

PR #1802, opened 2026-01-15
Hytale: Update Join/Leave/Chat Regex

winglessraven:
Update join/leave/chat regex for Hytale

Join now only captures after whitelist approval.
Leave changed to match same log format
Chat changed to avoid unintended captures

### DIFF
--- a/hytale.kvp
+++ b/hytale.kvp
@@ -100,9 +100,9 @@ Console.FilterMatchRegex=
 Console.FilterMatchReplacement=
 Console.ThrowawayMessageRegex=^(WARNING|ERROR): Shader.+$
 Console.AppReadyRegex=^\[[\d\/]+ [\d:]+\s+INFO\]\s+\[HytaleServer\]\s+Hytale Server Booted! \[Multiplayer(?:, Fresh Universe)?\] took .*$
-Console.UserJoinRegex=^\[[\d\/]+ [\d:]+\s+INFO\]\s+\[HandshakeHandler\]\s+Mutual authentication complete for\s+(?<username>.+?)\s+\((?<userid>.+?)\)\s+from\s+.+?\s+\(\/(?<endpoint>(?:\d{1,3}(?:\.\d{1,3}){3})|(?:[0-9a-fA-F:]+)):\d+,\s*streamId=\d+\)\s*$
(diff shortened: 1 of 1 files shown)


winglessraven:
> Thanks for looking at this.
> 
> The chat regex still matches lines it shouldn't. Given that Hytale usernames can only include numbers, letters, and underscores, maybe change it to exclude a space rather than colon? Seems to work when I tested
> 
> On the user join regex, I was reluctant to move it off the original line, since that captures IP as well. Maybe the leave regex can be expanded to pick up situations where a user is kicked for whitelist or password reasons?
> 
> This is probably not ideal as well since if, say, someone has discord notifications set up when a player joins, they will get notified of false positives.

OK, will take another look at the chat.

Main reason I used the original line was for the IP. Will double check if any of the lines showing the IP after the whitelist check could be suitable. Failing that your suggestion might be best, regardless of possible Discord notifications. I guess it depends what's more important, recording the IP or false notifications.

winglessraven:
In theory this should work for the leave event to pick up whitelist fail and the existing disconnect but AMP doesn't seem to remove the player when only the IP matches. Pretty sure that was a bug years back, thought it had been fixed.

New log line: 
```
[2026/01/15 16:46:21   INFO]                          [Hytale] Disconnecting QuicConnectionAddress{connId=9e3e5909eb67dbb1fc3a6ec1ac042d5dd4d9a766} (/84.65.40.38:49890, streamId=0) with the message: You are not whitelisted!
```

New regex: 
```
^\[[\d\/]+\s+[\d:]+\s+INFO\]\s+(?:\[Hytale\]\s+Disconnecting\s+QuicConnectionAddress\{connId=[0-9a-fA-F]+\}\s+\(\/(?<endpoint>(?:\d{1,3}(?:\.\d{1,3}){3})|(?:[0-9a-fA-F:]+)):\d+,\s*streamId=\d+\)\s+with the message:\s+'?You are not whitelisted!'?\s*|\[Universe\|P\]\s+Removing player\s+'(?<username>.+?)'\s+\((?<userid>[0-9a-fA-F]{8}(?:-[0-9a-fA-F]{4}){3}-[0-9a-fA-F]{12})\)\s*)$
```

winglessraven:
Only other line I can see that doesn't show up anywhere else is `[2026/01/15 16:46:21   INFO]                          [Hytale] {Setup(null (null, streamId=0)), winglessraven, 29fd62e4-c129-4c74-b62e-5a1ef8111619, SECURE} was closed.`

Can the config have multiple regex patterns to match on for one setting? Can't duplicate capture groups in one pattern so wondering how to handle it.

winglessraven:
OK, leave will now capture on
```
[2026/01/15 16:47:03   INFO] [Universe|P] Removing player 'winglessraven' (29fd62e4-c129-4c74-b62e-5a1ef8111619)
```
and
```
[2026/01/15 16:46:21   INFO] [Hytale] {Setup(null (null, streamId=0)), winglessraven, 29fd62e4-c129-4c74-b62e-5a1ef8111619, SECURE} was closed.
```

to handle the whitelist issue.

Chat should be a bit more robust aswell.

Greelan:
I'd prefer to be a little less strict with the username, to avoid having to update the regex if Hytale decide in future to allow other characters like a dash or whatever

PR #1806, opened 2026-01-19
Starrupture vulnerbility fix

AlienXAXS:
### Hello
I'm AlienX the developer behind the site https://starrupture-utilities.com/.

### What Is this For?
As described here: https://wiki.starrupture-utilities.com/en/dedicated-server/Vulnerability-Announcement there is a known vuln with any StarRupture server listening on TCP.

This allows complete remote control of the StarRupture server over the TCP port without any need for authentication, the wiki article goes into depth a little more without giving too much away.

### Mitigation
Adding the parameters as edited into the servers launch command will internally disable both the HTTP listener, and the WebSocket listner (both are vulnerable to attacks)

Also denying TCP by default will ensure no HTTP/WebSocket traffic will ever hit the server.

### Why Both?
While it is not in theory needed to implement the launch command when TCP is not forwarded, it makes sense to cover all bases.

### How Will AMP Server Owners Manage Their Server Now?
Thankfully as AMP already uses DSSettings.txt, users of AMP who start a StarRupture Dedicated Server have never been able to use the in game Server Manager.  This is by design.

### How To Reach Me

Any questions regarding this that require a private conversation, I can be reached on Discord under `alienx`

Thanks.


AlienXAXS:
Sorry for missing the version, was my first look at an AMP Template - ty for doing that for me!

AlienXAXS:
It's just RSA public-key encryption with PKCS#1 v1.5, then base64 encoded.

Greelan:
I was mainly interested in a way of users just generating the hash rather than the full json blob. AMP's settings just require the hash, but some users have mistakenly copied the whole json blob from your site

Greelan:
Ah, that option is perfect. Thanks. No need for an amp specific url, I will just update the description. 

What's your connection to the StarRupture devs?

PR #1832, opened 2026-02-22
Add Pumpkin

clovertortoise:
First timer here.

I put together a template for [Pumpkin](https://github.com/Pumpkin-MC/Pumpkin), which is a highly performant Rust implementation of a Minecraft server.

To get the config parsing right, I referenced how beammp handles TOML files and grabbed a few other tricks from existing templates to hack this together into a working setup.

It successfully deploys and runs on Linux / ARM64.

Notes:
* The Pumpkin server software itself is still in heavy active development, so some features are still a work in progress.
* I haven't had a chance to test on Windows yet.


pumpkin.kvp
```diff
@@ -0,0 +1,112 @@
+Meta.DisplayName=Pumpkin
+Meta.Description=The world's fastest Minecraft server. Pumpkin is in early development. Built with Rust for maximum performance.
+Meta.OS=3
+Meta.AarchSupport=Native
+Meta.Arch=All
+Meta.Author=PumpkinMC
```
Greelan: Should be your name

pumpkin.kvp
```diff
@@ -0,0 +1,112 @@
...
+Meta.DisplayImageSource=url:https://raw.githubusercontent.com/clovertortoise/AMPTemplates/pumpkin-template/pumpkinmc.png
+Meta.EndpointURIFormat={ip}:{GenericModule.App.Ports.$ServerPort}
+Meta.ConfigManifest=pumpkinconfig.json
+Meta.MetaConfigManifest=pumpkinmetaconfig.json
+Meta.ConfigRoot=pumpkin.kvp
+Meta.MinAMPVersion=2.4.0.0
+Meta.SpecificDockerImage=cubecoders/ampbase:debian
+Meta.DockerRequired=False
+Meta.DockerBaseReadOnly=False
+Meta.ContainerPolicy=NoPreference
```
Greelan: Change to `Supported`

pumpkinconfig.json
```diff
@@ -0,0 +1,1219 @@
+[
```
Greelan: As a general comment, please look at the other MC related templates (and the MC Java module) and adopt consistent setting headings, descriptions and ordering etc where applicable

Greelan (changes requested):
Very nice first attempt! It's looking very solid. I've suggested a number of tidies/improvements on a first pass

Greelan:
@clovertortoise is this finished yet? Has it been tested on Windows?

I note you haven't accepted AI's suggestion re double quotes for array values. Shouldn't at least Features.toml be made consistent?

clovertortoise:
> No worries, I will give it a proper going-over.
> 
> Would it make sense to call the Server Port instead Java Port? To align with/distinguish from Bedrock Port?


Much appreciated! Good catch on the port alignment, I’ll update that now.

Greelan:
@clovertortoise sorry for the delay in reviewing the updated template.

I have committed a few tidies (consistency etc). Have a look and let me know any issues.

A few other comments:
- `LAN Broadcast Port`: how does this relate to the port allocation in AMP?
- `Logging Color`: this should be irrelevant in the AMP context so you may want to default it to off and hide the setting.
- `Logging Level`: if this is set to off or a lower level than Info, will it impact the player and chat regex tracking?
- `Use Environment Log Level`: there is no corresponding setting in the toml files. Was this meant to set an env var in the KVP?
- `Use TTY`: this should be irrelevant in the AMP context?

Also, when I spun up a server on Linux, there were a few log messages that I thought worth checking with you as to whether they are issues:

```
features.toml changed because values were missing. The missing values were filled with default values.
...
2026-05-04 01:18:46  WARN main ThreadId(01) pumpkin::server: Failed to get level_info, using default instead: Info not found!
2026-05-04 01:18:46 ERROR main ThreadId(01) pumpkin::server: Failed to save level.dat: Info not found!
2026-05-04 01:18:46  WARN main ThreadId(01) pumpkin::server::connection_cache: Failed to load favicon from 'icon.png': not found; using default.
...
```

clovertortoise:
Hey @Greelan, no worries and thanks for the tidies, the changes look great on my end! To answer your questions:

- `LAN Broadcast Port`: If ever enabled, it would require an additional port allocation. It's similar to the enable-lan-visibility setting for bedrock servers, but allows setting the port manually.
- `Logging Color & Use TTY`: Agreed, both are irrelevant here. I've hidden them and defaulted them to false.
- `Logging Level`: Testing lower verbosity levels didn't appear to affect AMP's regex tracking.
- `Use Environment Log Level`: It was simply missing from the base features.toml file. This tells Pumpkin whether to accept RUST_LOG environment variables. It's properly synced up now.


Regarding the missing settings warning, if any options aren't in the base features.toml, the server binary sets their values to default and logs that warning.

As for the other startup log messages, they are harmless noise:

The level.dat and icon.png errors are the server complaining that the files don't exist yet (on a fresh install it falls back to the default icon and generates the world. Pumpkin still logs an error message for this right now).

I've pushed these adjustments and incorporated two new options: send recipes and plugin permissions. Let me know what you think.

Greelan:
Nice, will check it out. 

Re the LAN Broadcast Port, I suggest either adding that port pre-emptively in the template or at least including a note in the settings that the user will need to add the port to their instance manually to match. 

On Bedrock,enabling the LAN setting forces the server to use the default ports, which is a PITA. Is Pumpkin different? How do clients find the server on LAN anyway if it is using a non-default port?

Greelan:
One other thought: would it be better for the Category name to be `Pumpkin` rather than `Minecraft`?

clovertortoise:
My initial comparison to Bedrock was a bit of a red herring lol. It's sending UDP multicast. The LAN Broadcast setting in Pumpkin only affects Java clients.

Good idea on the user note, that seems like the most efficient route. I added a warning in the setting's description.

clovertortoise:
As for the Category name, I initially played around with that but opting for one Minecraft category felt more consistent. If more engine specific features get developed, we could definitely spin those out into a separate category, but the current state feels the most intuitive to me. Though, let me know if ya'll prefer a different design philosophy and we can easily switch it up.

Greelan (approved):

PR #1853, opened 2026-03-15
Add Archean template

Hedrauta:
Template is fully tested on Windows (non-container, fresh install on a VM) as well as Linux (default debian:ampbase container).

ContainerPolicy is set to ReccomendOnLinux, if trying to run with something different than Ubuntu 24+. 
Reason: Older Repos can have issues, bc glibc might be outdated.


IceOfWraith (approved):

IceOfWraith:
Thank you! I made some adjustments and got this merged. If you see any issues I can get them addressed as well.

Hedrauta:
I see, formatting and a few changes and additions. Thank you. 


2 Quick question : 

Isn't it smarter to use the `-port` argument to enforce the game's port? One bad timed change and users could use a different one. Just to be sure, yk 😉.

And why is the exit method on windows "none"? 🤔 It's the same for both.

IceOfWraith:
I couldn't find anything referencing a command line argument for the port. Do you have something that shows it? We usually put it there and in the file if it's possible on both just to avoid people doing something in one and messing up the server.

That Windows setting is an override for the other setting so it'll use "string" just like Linux. It was a work around setting added somewhat recently and when defaulted to "None" it doesn't do anything and falls back to the overall setting.

Hedrauta:
Oh... 

It's because there's no reference material... I pulled the strings out of the executables and searched myself. 

It works, I can ensure that. Using them for both of my archean servers. There are also a few more arguments, which I forgot 😅

IceOfWraith:
Understood. I noticed their lack of documentation as well. I decided to just trust you on the megalodon and game mode settings. x)

I'll get the command line adjusted to what you had!

Hedrauta:
Game mode was adjusted by a dev itself, and the megalodon-setting was an April's fool joke, which stayed inside the game as switchable setting.

PR #1880, opened 2026-04-15
Change game port from 7777 to 8777

rapid3642:
The official Astroneer site calls out that dedicated servers need to run on Port 8777, more information can be found here https://blog.astroneer.space/p/astroneer-dedicated-server-details/

I was not able to get the server running myself until I made this correction myself on my own AMP game server. This will help majorly with a ton of people running into these issues as stated in the Astroneer Discord server.

image

### DIFF
--- a/astroneerports.json
+++ b/astroneerports.json
@@ -1,7 +1,7 @@
 [
     {
         "Protocol": "UDP",
-        "Port": 7777,
+        "Port": 8777,
         "Ref": "ApplicationPort1",
         "Name": "Game Port",
         "Description": "Port for game traffic"


Greelan:
The default port when no parameter is set is 7777. The guide you linked only gives 8777 as an example. As it also says, you can use any port of your choosing.

rapid3642:
true, but fire up the server with 7777 then try it, i bet it wont work, then set it to 8777 and all of a sudden it works. trust me, 8777 is the fix :) even with port forwarding to 7777 it still wont work, 8777 does. you can also fire up the server instance and use https://astroservercheck.joejoetv.de/ to test if the server is in good health.

Greelan:
I don't really need to do so since I know it is fine.

But to humour you:

Screenshot 2026-04-15 at 10 59 26 am

rapid3642:
ah nice, thanks for humoring it, not sure what the issue when i set it up with the default 7777 didnt work for me until I set it to 8777. but glad it works for others.

Greelan:
Likely conflicting ports. Your screenshot shows you added a custom port on 7777.

PR #1910, opened 2026-05-19
Dune Awakening Retail update

PhonicUK:
Updated for the retail release of the server, supports full instance orchestration to support deep desert etc.


rumpel179:
Why is this version now usable only with Docker or Podman? It worked perfectly fine without them before. I really can't understand why we are now being forced to use Docker and similar tools. I have so many people here using AMP precisely because Docker isn't a mandatory requirement. It's a shame that it has been implemented this way. Still, thanks for your excellent work up to this point—we were at least able to host the game for a few days as it was. Bye!

IceOfWraith:
> Why is this version now usable only with Docker or Podman? It worked perfectly fine without them before. I really can't understand why we are now being forced to use Docker and similar tools. I have so many people here using AMP precisely because Docker isn't a mandatory requirement. It's a shame that it has been implemented this way. Still, thanks for your excellent work up to this point—we were at least able to host the game for a few days as it was. Bye!

What's the concern with using Docker or Podman? It's as easy as running `getamp installPodman` on Linux usually. This server is one where making it required does make sense because the ridiculous amount of dependencies.

rumpel179:
Sorry, but there are simply people who would prefer a "No Docker, No Podman" version. It’s not even about—as you put it—those few "ridiculous" dependencies. Why aren't we given the option to make it available *without* Docker, just like all the other instances? Given your level of expertise, surely it shouldn't be a problem to offer one version *with* Docker/Podman and another—just like the others—*without* it. If the old version had simply been left in place, that would have been perfectly fine. Best regards.

IceOfWraith:
You didn't really provide a reason. Docker and Podman are designed to work within AMP transparently so you shouldn't even know it's there. It eliminated the need for you to manually setup your server. I don't really see a downside. As for why on this template? Not sure. Mike set it up so he would know why. I suspect it needs to be isolated due to the multiple moving parts.

rumpel179:
One of the main reasons why I prefer to avoid Docker, Podman, and similar container-based solutions in the gaming/hosting environment is due to bad experiences in the past. While these systems may look flexible and modern at first, in practice they often introduced additional complexity and unnecessary maintenance overhead.

Especially with game servers or latency-sensitive applications, I repeatedly encountered performance and networking issues, higher resource usage, and complications related to updates, file permissions, backups, and troubleshooting. There were also recurring compatibility problems with mods, plugins, or certain custom server setups.

Because of this, I now generally prefer simpler and more direct solutions without an additional container layer, especially when stability, transparency, and ease of maintenance are the main priorities. bye

PhonicUK:
> Why is this version now usable only with Docker or Podman? It worked perfectly fine without them before. I really can't understand why we are now being forced to use Docker and similar tools. I have so many people here using AMP precisely because Docker isn't a mandatory requirement. It's a shame that it has been implemented this way. Still, thanks for your excellent work up to this point—we were at least able to host the game for a few days as it was. Bye!

@rumpel179 If it were possible without Docker or Podman I would have made it so. This is *just* for this game because of its unique requirements.

The problem is that the original server requires access to parts of the host system that are normally only accessible as *root* - so AMP would not be able to write the required information to those areas for the server process to read.

Inside a container, this isn't a problem because we can make an exception inside that container safely - but without it there's no way for AMP to give itself permission to write to those areas.

The particular areas that are required are also *security sensitive* - so not only is it not practical, it would be outright entirely unsafe to expose the particular files in question on a real host system.

The reason the *old* version worked without a container is because it lacked the orchestration layer, you couldn't visit the deep desert, Arrakeen, or any of the instances and encounters - you only had access to the Hagga Basin and the overmap. So most of the game was unplayable on the original non-Docker template.

Now you could manually remove the line that tells AMP it has to run in a container and it will start up, but you won't be able to play most of the game and a lot of the features won't work or behave correctly. If that's what you want to do and just stay in Hagga Basin then that's fine, but obviously that's not something that is reasonable to ship in general.

Requirements like this are *never* arbitrary, they are *always* carefully considered with regards to functionality, safety, security and performance.

PR #1911, opened 2026-05-21
Add Generic template for The Bus (German/English)

BroGamerYT:
This pull request adds a Generic module template for The Bus Dedicated Server.

Key Features:
- Supports both German and English localizations.
- Uses a Proton/Wine GE compatibility layer to run the Windows server binary on Linux hosts, solving the native Linux crash due to AVX2 requirements on older server CPUs.
- Fully tested and working with custom ports and Steam Workshop key configuration.


IceOfWraith:
Only English submissions should be submitted. There's a separate https://github.com/CubeCoders/AMPTranslations repo for other languages. I'll take a look at the template now.

IceOfWraith:
I can't find anything online regarding the AVX2 requirements or how Proton fixes that. Can you share where you're seeing it? I suspect my search results are just bad since "the bus" is a pretty generic search term.

IceOfWraith:
Closing due to signs of using either the config gen or AI to generate this one. If you clean it up using the new version of the config gen it may be accepted. But the answers above would need addressed still.

PR #1940, opened 2026-07-10
Fix player regex & add 2 new convars

killaketa:
Fix player join, leave, & chat regex so player count works properly. Also added 2 ConVars to the config. g_friendlySaber should've been included alongside g_friendlyFire as it is essentially friendlyFire but for lightsabers. g_privateduel enables the ability for 2 players to 1v1 duel each other.


killaketa:
Also changed g_allowvote into a number input on the config since its a bitflag in OpenJK (the server used for JKA on AMP) and modern forks.

PR #1941, opened 2026-07-11
Draft: Add The Riftbreaker Dedicated server template

CostcoFanboy:
Adds an AMP Generic module template for The Riftbreaker dedicated server (Steam app 4114030). Supports Windows and Linux (wine for Linux).

Made it work on Windows, not tested yet in Linux hence draft.


riftbreakerwinescript.sh
CostcoFanboy, line 1: Follows the pattern I was in v-rising and space-engineers.

Greelan:
Why reinvent the wheel?

PR #1945, opened 2026-07-22
Update Palworld to include a multi-threading suggestion for 1.0

klopsknoedel:
This adds a hint to the multi-threading settings to warn about potential performance degradation with Palworld 1.0+
I've tested this myself and enabling multi-threading in v1 may cause some issues. For example base/camp-inventories can get out of sync. If you use some materials at a workbench, the consumption may not be reflected to the client where you "have" the exact same amount of materials as before crafting.
Also rubber banding can be slightly worse and other sync/concurrency issues can occur.

So, at least, users should be warned about turning this setting on for Palworld 1.0

This also reflects the official server arguments description: https://docs.palworldgame.com/settings-and-operation/arguments/#regarding-performances


IceOfWraith:
Thank you!

PR #1946, opened 2026-07-29
Update wurm-unlimitedports.json

unixbird:
Adjusted as this is the main port where clients connect per: https://www.wurmpedia.com/index.php/Server_administration_(Wurm_Unlimited)

3724 TCP should be the externalport rather than the 8766 UDP that was originally specified


unixbird:
Added another commit to fix it so the analytics work properly

IceOfWraith:
The port changes look good, but I'm not seeing logging come through for the user join/leave that you added. Can you verify where you saw those lines at?

unixbird:
> The port changes look good, but I'm not seeing logging come through for the user join/leave that you added. Can you verify where you saw those lines at?

I saw them in the console when someone would join and leave (or disconnect as the message is about the same) and it seems to work fine on my end in that regard as I get data

IceOfWraith:
Are you on Windows or Linux? Do you have any logging settings you might have adjusted?

unixbird:
> Are you on Windows or Linux? Do you have any logging settings you might have adjusted?

I'm on linux 

the console will have entries like this from my experience: 

`[06:49:33 AM] INFO com.wurmonline.server.LoginHandler: /9.9.9.9,Ken successfully logged on, id: 99999999.
[07:10:28 AM] INFO com.wurmonline.server.LoginHandler: /8.8.8.8,David successfully reconnected.
[07:02:53 AM] INFO com.wurmonline.server.players.Player: Logout complete for Player [id: 99999999999, name: Ken]`

anonymized the entries a little but this is what i see

unixbird:
@IceOfWraith found out its actually from the AnnouncerMod 

From my understanding this is a pretty core mod that many would be running on their server rather than vanilla anyway so I think it wouldn't hurt to keep it as is and merge it

IceOfWraith:
I was debating how to handle this one then got distracted with code. I'm looking through the pull requests again and I think the only way I'm okay with adding that in is if the mod is included in the template as an option as well. I don't mind adding that part in though. Otherwise, people will have scheduler tasks and options that look like player tracking should work when it doesn't by default.

IceOfWraith:
I do want the port fixed right away though so I'll apply that part of the change manually to the repo and we can address the player tracking here separately.

unixbird:
> I was debating how to handle this one then got distracted with code. I'm looking through the pull requests again and I think the only way I'm okay with adding that in is if the mod is included in the template as an option as well. I don't mind adding that part in though. Otherwise, people will have scheduler tasks and options that look like player tracking should work when it doesn't by default.

That's fair. I would say that no one is running Vanilla wurm these days. To run any mods it needs a patched client: https://github.com/ago1024/WurmServerModLauncher

Announcer is the mod that gives us this log in/log out information


Seeing as Wurm unlimited isn't ever gonna update it wouldn't be terrible to possibly have that as the default

IceOfWraith:
@unixbird I've added the automatic download handling for the mod loader and Announcer. If it all looks good to you I can get this merged.

IceOfWraith:
I'm going to go ahead and merge this and if there's any residual issues let me know and I can address them.

PR #1950, opened 2026-08-03
Ark SA template add Cryopod related settings

BlackLamb:
Added settings for the Cryopod and Cryofridge that allow them to act closer to how Ark SE works. The three settings are:

- Disabling the requirement for a Cryofridge to use a Cryopod 
- Disabling the enemy check for deploying Cryopods
- Allowing Cryofridges on saddles and rafts


IceOfWraith:
Linking this to another pull request so they both can be merged together. https://github.com/CubeCoders/AMPTemplates/pull/1831

PR #1951, opened 2026-08-03
Add proper UserJoin, UserLeave, and UserChat regex alongside important ConVar fix

killaketa:
The UserJoin/Leave/Chat regex in Jedi Academy was completely empty so I made some to properly account for players and bots. Players have a choice of joining with no GUID which passes a GUID of `NOGUID` to the server. There is a dropdown list added to the server config which allows hosters to either ignore or account for bots and NOGUID players. 

The `g_allowVote` convar was described and set incorrectly. It is actually a bitflag which allows the hoster to enable/disable certain aspects of voting (see [OpenJK CallVote function](https://github.com/JACoders/OpenJK/blob/1a6a643427aa347553e9073dac5570b33337c4d9/codemp/game/g_cmds.c#L2139)). Ive also added the `g_privateduel` and `g_friendlysaber` convars into the config which enable private dueling between players & enable friendly fire damage with lightsabers.


IceOfWraith:
Thank you! I've made some tweaks to the changes you proposed. If they look okay I'll get this merged.

killaketa:
Thank you for the convar & description updates but the regex should be reverted to ignore the digit and player endpoint, occasionally the server doesn't print them in certain disconnect messages. Keeping it a wildcard stops the regex from breaking in those rare scenarios. The only real reliable way to track players is the OpenJK GUID (granted the player didn't disable it).

IceOfWraith:
Understood. Got that change reverted!

IceOfWraith (approved):

PR #1953, opened 2026-08-06
Update FiveM, RedM, and txAdmin Templates

Sethy1124:
Don't mind the messy commits :P

**Changes:**
Global
- Updated AppReadyRegex. Cfx Nucleus is being deprecated.
- Changed the Linux Executable to the run.sh file provided in the server artifacts.
- Disabled SleepMode. Cfx servers can take a significant amount of time to start, so sleeping seems impractical.

FiveM & RedM
- Changed ExitMethod. quit causes the server to restart rather than shutdown
- Reorganized server.cfg. The main issue was access control being below the resources. This will cause issues with ace commands not being applied to their respective scripts.
- Removed serverProfile from the command line args. Simply not needed since it was set to default.

FiveM
 - Clarified GTA V Legacy in Title and Description
 - Added New Game Build.

RedM

txAdmin
- Removed the Environmental Variables from the CommandLineArgs. The variables that were used are outdated.

Another issue is that the RCON in FiveM and RedM is not working. The server returns that the RCON password is not set. Adding it to the server.cfg would work; however, I was not sure how to implement the password being mapped to the config.


IceOfWraith:
I started testing this against FiveM first but I don't see that startup message. I was trying against the "Latest" release. Is there one you were testing against I can mirror on my side?

Sethy1124:
Yes, you are correct. That message is only present if txAdmin is being used, silly me. I suppose the next best option for txAdmin-less servers would be checking that the license key authenticated.


Ideally, in my opinion, the FiveM and RedM templates should be deprecated entirely. While it may seem counterintuitive to have two "server managers" overlapping, txAdmin can do a lot that AMP cannot.

IceOfWraith:
I reverted a few of the changes that were based on opinions rather than functionality. And tweaked a few others further. Please let me know if you see any issues otherwise we can get this merged.

I switched the name back to remove Legacy since there's an early access build of Enhanced FiveM which I'd like to see added to this template eventually.

For future reference, we don't typically allow using a script as an executable aside form rare cases. The proper way to handle it is running the commands/setup from the script within AMP's settings like it already was.

txadmin.kvp
```diff
@@ -22,20 +22,20 @@ Meta.ExtraContainerPackages=[]
...
-App.ExecutableWin=server/FXServer.exe
-App.ExecutableLinux=server/run.sh
+App.ExecutableWin=server\FXServer.exe
+App.ExecutableLinux=server/alpine/opt/cfx-server/ld-musl-x86_64.so.1
 App.WorkingDir=server-data
-App.LinuxCommandLineArgs=
+App.LinuxCommandLineArgs=--library-path "{{$FullBaseDir}}server/alpine/usr/lib/v8/:{{$FullBaseDir}}server/alpine/lib/:{{$FullBaseDir}}server/alpine/usr/lib/" -- "{{$FullBaseDir}}server/alpine/opt/cfx-server/FXServer" +set citizen_dir "{{$FullBaseDir}}server/alpine/opt/cfx-server/citizen/"
 App.WindowsCommandLineArgs=
-App.CommandLineArgs={{$PlatformArgs}}
+App.CommandLineArgs={{$PlatformArgs}} +set txAdminPort {{$txAdminPort}} +set txAdminInterface {{$ApplicationIPBinding}}
```
Sethy1124: According to the txAdmin docs. These convars are deprecated. [Source](https://github.com/citizenfx/txAdmin/blob/master/docs/env-config.md)

IceOfWraith: You're absolutely right. Not sure how I didn't catch that in the documents.

IceOfWraith:
I've gone ahead and merged this one since it's impacting so many people. But if you see anything else off let me know and I can merge it directly.

Sethy1124:
Everything should be good for now. I'll see about adding Enhanced in the future, likely once asset escrow has been implemented. 

Thanks for the help and info. I'll have a cleaner commit list next time :)

IceOfWraith:
No, it was good!

PR #1956, opened 2026-08-10
Update project-zomboidmetaconfig.json

alelejo:
Added config to allow AMP to edit {{servername}}_SandboxVars.lua file like it was editing the {{servername}}.ini file


IceOfWraith:
I see that you added the metaconfig side but from a spot check I don't see that any settings are configured that exist in the lua file.

alelejo:
That's the thing this PR fixes, and I have a before/after for it.

project-zomboidconfig.json already declares settings whose ParamFieldName also exists as a top-level key in {servername}_SandboxVars.lua — MinutesPerPage, HoursForLootRespawn, MaxItemsForLootRespawn among them. On an unmodified b42 instance I changed MinutesPerPage from the UI and it updated the .ini only; the .lua was untouched, because there was no ConfigFile entry for it. After adding the metaconfig entry, the same change reaches both files.

So those settings are already silently not applying today. That's what the entry addresses.

It only works partially, though.

1. A key that exists in both files gets written to both. For example `Map` is a plain value in the .ini and a nested table opener (`Map = {`) in the .lua — AutoMap overwrites the opener and corrupts the file.

2. My workaround is a lookahead in ConfigFormatRegex that skips any line whose value starts with `{`, so table openers are never writable

3. This protects the table structure but does not scope nested scalars — a key like ZombieLore.Speed still matches as a bare `Speed`. No current setting collides, but it's luck rather than design.

IceOfWraith:
Gotcha. Let me look at it closer with those settings in mind then and see if this can be accepted.

IceOfWraith:
Okay, I see what's going on with it now. The devs moved a few of the settings to the lua file but also changed the names of some other settings. I'm going to make some commits that include all the changes but keep backwards compatibility soon.

alelejo:
Understood, thanks for the help.

IceOfWraith:
@alelejo It actually looked good. I made a few small tweaks to the settings just to account for other changes, but nothing that would have prevented your change from working. Let me know if you see any issues, otherwise we can get this merged.

alelejo (commented):
All good from my side

PR #1957, opened 2026-08-10
Add Mount And Blade 2: Bannerlord Co-op 

ronanmcdonad-source:


IceOfWraith:
Templates made with the config generator aren't accepted without major refactoring. This config also wouldn't work properly.

PR #1967, opened 2026-08-23
Updated 4.0 installation and config and created 4.1 configs

Trinagan:
For SPT 4.0:
- Changed the `updates.json` to target the correct downloads for SPT and Fika files for 4.0
- Added the revive settings to `config.json` that were added

For SPT 4.1:
- Created new set of configs due to folder structure change and project moving to a new repo
- Set the default .NET version to 10.0 as that's what SPT 4.1 uses
- Adjusted the regex to match the new release URL
- Introduced the same new config settings for revive present in 4.0


IceOfWraith:
Thank you for the submission! I've made one commit with further tweaks if you wouldn't mind verifying if I missed anything. It looks like the server downloads and runs though. Only real blocker on Windows was .NET 9 still being specified in the update stages. Everything else was just adding new settings and matching the default.

Trinagan:
Thanks for the revision - I have indeed missed the Windows .NET version as I use Linux for hosting and did not have a setup to test it on Windows.
Everything else looks correct as far as I can tell.

IceOfWraith (approved):

PR #1972, opened 2026-08-27
Update project-zomboid.kvp

HuguesDePayens:
Project Zomboid: detect Workshop mod updates from console

Set Console.UpdateAvailableRegex to detect the
"CheckModsNeedUpdate: Mods need update" response.

This allows AMP to emit an update-available event when Project Zomboid detects that one or more configured Workshop mods need updating, making it possible to trigger automated restart/update workflows through the scheduler.


Greelan:
From recollection, that regex isn't actually wired to anything :)

HuguesDePayens:
Thanks, that explains why I couldn’t get it to trigger anything.

The use case I’m trying to solve is Project Zomboid Workshop mod updates. PZ has a built-in console command:

checkModsNeedUpdate

which asynchronously outputs:

CheckModsNeedUpdate: Mods need update

when one or more Workshop mods require an update.

AMP’s scheduler can send the command, but currently there doesn’t appear to be a way to react to the resulting console output.

Would it make sense to wire Console.UpdateAvailableRegex into GenericModule so a match emits an update-available scheduler event?

Alternatively, is there another GenericModule mechanism for triggering a scheduler event from a console regex that I’m overlooking?

This would allow PZ admins to periodically run checkModsNeedUpdate, then warn players, back up, and restart only when a Workshop mod actually needs updating.

IceOfWraith:
Not currently. It's a planned feature though.

HuguesDePayens:
This is a pretty needed feature. It makes it so that a manual restart is constantly required whenever someone tries to login. Make modding almost pointless for Project Zomboid :( I hope there is a fix for it soon. Thanks for all your hard work!

PR #1973, opened 2026-08-29
Add RimWorld Multiplayer (Zetrith) Server

luna61ouo:
Adds a template for the standalone dedicated server of **Zetrith's Multiplayer** (`rwmt/Multiplayer`).

### Credit

The mod and the dedicated server are the work of **Zetrith** and the [rwmt/Multiplayer](https://github.com/rwmt/Multiplayer) contributors. This PR contains none of their code — it only teaches AMP how to download and run their published release, so `Meta.URL` points at their repository and the description makes clear that the template is just packaging. `Meta.Author` follows the convention used throughout this repo and names the template author.

This is a different project from the existing `rimworld-together` template — different mod, different server, different protocol. Players need the matching client mod from the same release.

### Submission checklist

- **From-scratch setup from ADS** — done, repeatedly. Instance created from the app list, `Update Application` installs a self-contained .NET 8 runtime plus the server, `Start` brings it up, AMP reports **Ready**, console output and player count both work. Bootstrapped a real world through a game client and played on it.
- **All declared platforms verified** — declared `Meta.OS=Linux`, `Meta.Arch=x86_64`, `Meta.AarchSupport=Unknown`, because that is the only combination I have actually run. The Windows and arm64 stages are present but inert under this declaration; happy to widen it once someone has tested them.
- **Configuration manifest** — yes, three settings (release tag, .NET runtime version, server port).
- **Configuration meta-manifest** — **not provided**, see below.

### Two things a reviewer will want to know

**1. The server needs a real TTY, so the template installs a small launcher.**

The server's main loop polls `Console.KeyAvailable`. With `App.UseLinuxIOREDIR=False` AMP still gives it a pipe, so it throws about one second after start:

```
Unhandled exception. System.InvalidOperationException: Cannot see if a key has been pressed
when either application does not have a console or when console input has been redirected
from a file. Try Console.In.Peek.
   at System.Console.get_KeyAvailable()
   at Program.<Main>$(String[] args) in .../Source/Server/Server.cs:line 122
```

With "restart on unexpected stop" enabled this becomes an endless restart loop.

The install stage therefore renames the real binary to `Server.bin` and writes a three-line `Server` launcher that execs it under `script(1)` to give it a pty. `App.ExecutableLinux` stays `Server`, so nothing else in the template is affected and the launcher can simply be dropped if upstream ever removes the TTY assumption. `/usr/bin/script` (util-linux) is present in `cubecoders/ampbase:debian`.

Because the process then runs on a pty its output is CRLF-terminated, so the `Console.*` regexes are anchored with `\r?$`.

**2. No meta-manifest, because the server's config file is TOML.**

The server keeps its settings in `settings.toml`. `ConfigType: toml` does not appear to be supported — AMP answers `No meta configuration manifest data is available`, and no template in this repo uses it (58 ini, 36 json, 10 xml, 0 toml). So the meta-manifest was removed rather than shipped broken.

The practical consequence is that **AMP cannot manage the game port**: `settings.toml` stores it as a single merged `directAddress = "0.0.0.0:30502"` string rather than separate host/port fields, so the `$ServerPort` field is declared but hidden and the server binds whatever its own config says. This is documented in that setting's description. If TOML support exists under another name, or if you would prefer a pre-start stage that rewrites `directAddress` from `{{$ServerPort}}`, I am glad to add it.

### Other notes

- The upstream release ships one zip containing both `Server/Linux/` and `Server/Windows/`, so the install stage flattens the right one into the base directory. Note the Linux payload contains a file literally named `Server`, which collides with the extracted `Server/` directory — the stage moves the directory aside first.
- `continuous` is the default release tag; it is the rolling build and currently the only one that ships a server.
- The server intentionally shuts itself down once after a client finishes uploading the initial world, so "restart on unexpected stop" should be enabled.


IceOfWraith:
AI generated templates are not accepted into the official repo.

PR #1976, opened 2026-08-31
Update wreckfest2config.json

bkwar201:
Change Proton DefaultValue from "" to "GE-Proton10-34" At least the current version of Proton 11 downloaded it prevents the server from properly launching, not visible in game server browser nor direct connectable. No AMP logs reflecting a successful/failed server start. It does work with 10-34 when testing, still nothing shows up in AMP logs, must be a wreckfest2 thing.

### DIFF
--- a/wreckfest2config.json
+++ b/wreckfest2config.json
@@ -56,7 +56,7 @@
         "FieldName": "SpecificReleaseVersion",
         "InputType": "text",
         "ParamFieldName": "SpecificReleaseVersion",
-        "DefaultValue": "",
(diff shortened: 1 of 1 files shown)


IceOfWraith:
I was hoping they'd fix Proton before now. There's a number of games that aren't working on Proton 11. I don't want to set a default just yet since people won't realize it's hard set to that. I'd prefer they make a conscious change to their setting so they know to remove it later.

IceOfWraith:
For anyone that ends up here... You can edit the version under Configuration->Updates then click Update inside the instance.

Issue #1982, opened 2026-09-07
Tarkov SPT: changed release URL in sp-tushonka mirror


MoInMaRvZ:
the latest release found for SPT 4.1.0+ is this one
- https://mirror.sp-tushonka.com/releases/SPT-4.1.3-40743-ddce41c.7z

But there are two more versions available under another path
- https://mirror.sp-tushonka.com/builds/SPT-4.1.5-40743-7d7add5.7z
- https://mirror.sp-tushonka.com/builds/SPT-4.1.4-40743-072d534.7z

PR #1983, opened 2026-09-07
fixed tarkov spt versioning problem #1982

MoInMaRvZ:


IceOfWraith:
@MoInMaRvZ Thanks for the contribution! I've made some further changes to line it up with other templates and allow for empty = latest version. Please let me know if you see any issues with the changes and we can get this merged.

Trinagan:
I believe introducing this might cause issues to users - SPT does not keep old releases available for download as it uses Github LFS and cannot keep the older versions available.
Even right now, the only versions of SPT 4.1 that are available for download are 4.1.5, 4.1.4 and 4.1.2 from the archived old project.

IceOfWraith:
I guess I'm confused. It sounds like they do keep three versions at any time. Adding the version but defaulting to empty means the latest is always grabbed but if there's a bad release and someone wants to switch, they are able to.

Trinagan:
To hopefully clear up the confusion - the only reason 4.1.2 is available is because it was the last release before the old developers archived the project.
Since around 3.10 I believe SPT has not been holding onto older release versions as they were unable to keep the files available, and the moderation team is not big enough to be able to support many versions either.
The reason why 4.1.4 is still available is unknown to me, but usually it is only the latest version that is kept available, and people are always encouraged to update to the latest.

IceOfWraith:
I still think it's good to have the option to select the version based on that. Worst case someone puts in an older version that doesn't work and they have to remove it. Best case it's left empty as the description says and it downloads the latest.

IceOfWraith:
I'm going to go ahead and merge this one. If there's any issues found with the actual version handling let me know and I'll address them.

PR #1984, opened 2026-09-07
Switch to CDN

IceOfWraith:
Driven by the unexpected change from Microsoft: https://github.com/orgs/community/discussions/206581#discussioncomment-18269083

CubeCoders will be mirroring this repo every 5 minutes to a CloudFlare CDN for high availability. Retargeting all templates to use that as the source of files going forward.

One caveat is this change will not allow any `.config` files going forward. They should be named `.cfg` in the repo and renamed upon download.


PhonicUK (approved):
URLs are correct and config->cfg works as-is because UpdateSourceArgs keeps the original name.

PR #1991, opened 2026-09-16
Fix Valheim not saving when stopping the instance and also allow force save from AMP panel

Ludska01:
Changed the value of HasWriteableConsole to true
 to allow force save from the AMP console.
 
Changed the value of App.ExitMethod
 to allow the server to exit gracefully when the
 console is closed instead of failing to save the world correctly and stopping at first step.
image

Changed App.ExitTimeout to allow the server time
 to save bigger worlds before exiting.


dreamwraith:
seems like this should become a configurable value if it needs to change only for large worlds.

IceOfWraith:
This is actually a bug in AMP that's fixed in the coming release.

